PR #150, opened 2021-01-27
Updated homepage footer from generic template

growlix:
Updated homepage footer to say "Copyright © 2021 Facebook Inc." instead of "Copyright © 2021 Your Name or Your Company Name"


prigoyal:
looks like the fix already landed :) thank you @growlix for jumping on this and quickly fixing it :)

PR #162, opened 2021-01-29
[fix] missing import referenced in __all__ of "vissl.data"

QuentinDuval:
The function `get_data_files`  is referenced in `__all__`, but is not part of the imports above. The two valid choices are either to remove it from `__all__` or to import it. I thought the second option made more sense here but I might be wrong.

### DIFF
--- a/vissl/data/__init__.py
+++ b/vissl/data/__init__.py
@@ -8,7 +8,7 @@ from torch.utils.data import DataLoader
 from vissl.data.collators import get_collator
 from vissl.data.data_helper import StatefulDistributedSampler
 from vissl.data.dataloader_sync_gpu_wrapper import DataloaderSyncGPUWrapper
-from vissl.data.dataset_catalog import VisslDatasetCatalog, register_datasets
+from vissl.data.dataset_catalog import VisslDatasetCatalog, register_datasets, get_data_files
 from vissl.data.disk_dataset import DiskImageDataset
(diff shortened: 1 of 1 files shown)


prigoyal:
looks great @QuentinDuval , can you click on Import to Phabricator and we will run some integration tests and land the PR :D

prigoyal (approved):

vissl/data/__init__.py
```diff
@@ -8,7 +8,7 @@
 from vissl.data.collators import get_collator
 from vissl.data.data_helper import StatefulDistributedSampler
 from vissl.data.dataloader_sync_gpu_wrapper import DataloaderSyncGPUWrapper
-from vissl.data.dataset_catalog import VisslDatasetCatalog, register_datasets
+from vissl.data.dataset_catalog import VisslDatasetCatalog, register_datasets, get_data_files
```
prigoyal, line 11: small linter suggestion and then good to go! thank you @QuentinDuval :)
```
from vissl.data.dataset_catalog import (
    VisslDatasetCatalog,
    register_datasets,
    get_data_files,
)
```

QuentinDuval: Indeed :) It should be fixed in 0ec673d787a7ac959c99273a0c0e630f29184faa.

prigoyal:
thank you @QuentinDuval , we need to "reimport the pull request" and we can merge it right away afterwards :)

PR #163, opened 2021-01-29
Update configuration for the linear evaluation on place205 to use "disk_folder" dataset loader instead of "everstore"

QuentinDuval:
Currently, running the linear evaluation benchmark on places205 raises an error: "everstore" is not supported. Using "disk_folder" and referencing a places205_folder structures like the following solves the issue:

```
train/
   abbey/
      <images>
   ...
val/
   abbey/
      <images>
   ...
```


prigoyal (approved):
looks great @QuentinDuval , can you click on Import to Phabricator and we will run some integration tests and land the PR :D

PR #165, opened 2021-02-02
First draft for enriching the datasets of VISSL with Pytorch dataset that do not follow the "disk_folder" layout.

QuentinDuval:
This PR is a first draft to discuss the approach to integrate these datasets:
- is the "pytorch_dataset" a good idea for integrating (knowing that there might be overlaps, like "Imagenet" or "Places365")
- the hyper-parameters of the linear evaluation on both CIFAR10 and CIFAR100 are not tuned
- more datasets could be integrated this way: for instance MNIST (which is used as benchmark in the CLIP paper) but with modifications (we would have to wrap it to return images with 3 channels instead of 1 and of size 32)


prigoyal:
thank you @QuentinDuval , the PR looks great to me. I think it;s a very relevant question if we should keep the pytorch_dataset especially it leads to clash with imagenet and places. lmk think about this and get back to you

PR #179, opened 2021-02-08
Bug fix in the MLP layer: torch.squeeze was removing the batch size when equal to 1

QuentinDuval:
The fix comes with unit tests so that error will be caught up if it comes back.


prigoyal (approved):
thank you so so much and yay for the tests!

let's import to phabricator and merge it :)

PR #180, opened 2021-02-08
Support for torchvision datasets in VISSL (take 2)

QuentinDuval:
This PR is a revision of the first PR (https://github.com/facebookresearch/vissl/pull/165) which was a draft to discuss the approach to integrate these datasets. Compared to the previous PR:

1. More pytorch datasets are included (CIFAR10, CIFAR100, STL10, MNIST)
2. The different datasets are proposed in benchmark/linear_image_classification/datasets

To combine the datasets with the standard linear evaluation benchmark, we can use the following hydra syntax

```
python tools/run_distributed_engines.py config=benchmark/linear_image_classification/imagenet1k/eval_resnet_8gpu_transfer_in1k_linear config.MODEL.WEIGHTS_INIT.PARAMS_FILE=PATH +config/benchmark/linear_image_classification/datasets=cifar10
```

Note that for MNIST, a wrapper dataset was needed to convert the images from 1 channel to 3 channels, and increase their size to 32x32 to at least match the downsampling by 32 of standard Resnet50.

### DIFF
--- a/configs/config/benchmark/linear_image_classification/datasets/cifar10.yaml
+++ b/configs/config/benchmark/linear_image_classification/datasets/cifar10.yaml
@@ -1,0 +1,47 @@
+# @package _global_
+config:
+  DATA:
+    NUM_DATALOADER_WORKERS: 4
+    TRAIN:
+      DATA_SOURCES: [torchvision_dataset]
+      LABEL_SOURCES: [torchvision_dataset]
+      DATASET_NAMES: [CIFAR10]
+      BATCHSIZE_PER_REPLICA: 256
+      TRANSFORMS:
+        - name: RandomResizedCrop
+          size: 32
+        - name: RandomHorizontalFlip
+        - name: ToTensor
+        - name: Normalize
+          mean: [0.485, 0.456, 0.406]
+          std: [0.229, 0.224, 0.225]
+      MMAP_MODE: True
+      COPY_TO_LOCAL_DISK: False
+      COPY_DESTINATION_DIR: /tmp/cifar10/
+    TEST:
+      DATA_SOURCES: [torchvision_dataset]
+      LABEL_SOURCES: [torchvision_dataset]
+      DATASET_NAMES: [CIFAR10]
+      BATCHSIZE_PER_REPLICA: 256
+      TRANSFORMS:
+        - name: ToTensor
+        - name: Normalize
+          mean: [0.485, 0.456, 0.406]
+          std: [0.229, 0.224, 0.225]
+      MMAP_MODE: True
+      COPY_TO_LOCAL_DISK: False
+      COPY_DESTINATION_DIR: /tmp/cifar10/
+  METERS:
+    name: accuracy_list_meter
+    accuracy_list_meter:
+      num_meters: 1
+      topk_values: [1]
+  MODEL:
+    FEATURE_EVAL_SETTINGS:
+      LINEAR_EVAL_FEAT_POOL_OPS_MAP: [
+        ["res5", ["AdaptiveAvgPool2d", [[1, 1]]]],
+      ]
+    HEAD:
+      PARAMS: [
+        ["eval_mlp", {"in_channels": 2048, "dims": [2048, 10]}],
+      ]
--- a/configs/config/dataset_catalog.json
+++ b/configs/config/dataset_catalog.json
@@ -15,6 +15,22 @@
         "train": ["<img_path>", "<lbl_path>"],
         "val": ["<img_path>", "<lbl_path>"]
     },
+    "CIFAR10": {
+        "train": ["<path_to_pytorch_dataset>", "<unused>"],
+        "val": ["<path_to_pytorch_dataset>", "<unused>"]
+    },
+    "CIFAR100": {
+        "train": ["<path_to_pytorch_dataset>", "<unused>"],
+        "val": ["<path_to_pytorch_dataset>", "<unused>"]
+    },
+    "STL10": {
+        "train": ["<path_to_pytorch_dataset>", "<unused>"],
+        "val": ["<path_to_pytorch_dataset>", "<unused>"]
+    },
+    "MNIST": {
+        "train": ["<path_to_pytorch_dataset>", "<unused>"],
+        "val": ["<path_to_pytorch_dataset>", "<unused>"]
+    },
     "inaturalist2018_folder": {
         "train": ["<img_path>", "<lbl_path>"],
         "val": ["<img_path>", "<lbl_path>"]
--- a/vissl/data/ssl_transforms/img_pil_to_rgb_mode.py
+++ b/vissl/data/ssl_transforms/img_pil_to_rgb_mode.py
@@ -1,0 +1,59 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+from typing import Any, Dict, List, Union
+
+from PIL import Image
+from classy_vision.dataset.transforms import register_transform
+from classy_vision.dataset.transforms.classy_transform import ClassyTransform
+
+
+@register_transform("ImgPil2RGB")
+class ImgPil2RGB(ClassyTransform):
+    """
+    Convert a PIL image to RGB mode.
+
+    This transform is necessary to adapt datasets such as MNIST (which has 1 color channel)
+    as input to traditional architectures (like ResNet50) requiring 3 color channels.
+
+    We first create an image of the same size as the original (or fixed "size" if provided as input) and
+    paste the original image in it. If a different size is provided, the "box" indicates where the paste
+    of the original image will be in the target image.
+
+    Default behavior (no arguments provided) is to keep the original image size.
+
+    The output is a PIL image in RGB mode.
+    """
+
+    def __init__(self, size: Union[int, List[int]], box: List[int]):
+        super().__init__()
+        if isinstance(size, int):
+            size = [size, size]
+        if size:
+            assert len(size) == 2, f"ImgPil2RGB: Expected \"size\" parameter of length 2, got: {size}"
+        assert len(box) == 2, f"ImgPil2RGB: Expected \"box\" parameter of length 2, got: {box}"
+        self.size = size
+        self.box = box
+
+    def __call__(self, original_img: Image.Image) -> Image.Image:
+        if self.size:
+            img = Image.new(mode="RGB", size=self.size)
+            img.paste(original_img, box=self.box)
+        else:
+            img = Image.new(mode="RGB", size=original_img.size)
+            img.paste(original_img)
+        return img
+
+    @classmethod
+    def from_config(cls, config: Dict[str, Any]) -> "ImgPil2RGB":
+        """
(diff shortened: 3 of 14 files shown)


vissl/data/torchvision_dataset.py
```diff
@@ -0,0 +1,110 @@
...
+            return CIFAR10(self.path, train=is_train_split)
+        elif self.dataset_name == "CIFAR100":
+            return CIFAR100(self.path, train=is_train_split)
+        elif self.dataset_name == "STL10":
+            stl_split = "train" if is_train_split else "test"
+            return STL10(self.path, split=stl_split)
+        elif self.dataset_name == "MNIST":
+            return MNISTAdapter(root=self.path, train=is_train_split)
+        else:
+            raise ValueError(f"Unsupported dataset {self.dataset_name: str}")
```
prigoyal, line 50: we should convert this to an enum following example https://github.com/facebookresearch/vissl/blob/master/vissl/models/model_helpers.py#L239-L256 :)

vissl/data/torchvision_dataset.py
```diff
@@ -0,0 +1,110 @@
...
+        return image, is_success
+
+    def get_labels(self) -> List[int]:
+        """
+        Return the labels for each sample
+        """
+        return [self.dataset[i][1] for i in range(self.num_samples())]
+
+
+class MNISTAdapter(MNIST):
```
prigoyal, line 81: is this something that makes sense to implement as a transform for mnist only? that way we can keep the data transformation logic to transformations only. see some transformations like https://github.com/facebookresearch/vissl/blob/master/vissl/data/ssl_transforms/img_pil_to_lab_tensor.py where we convert images to LAB format. I would highly encourage to move this to transform the more I look at it unless it won't fit the bill :)

configs/config/dataset_catalog.json
```diff
@@ -15,6 +15,22 @@
         "train": ["<img_path>", "<lbl_path>"],
         "val": ["<img_path>", "<lbl_path>"]
     },
+    "CIFAR10": {
```
prigoyal, line 18: I love it!!! :)

configs/config/benchmark/linear_image_classification/datasets/cifar10.yaml
```diff
@@ -0,0 +1,47 @@
+# @package _global_
```
prigoyal, line 1: for all the config files, I am fine if we want to move them out to `linear_image_classification/<dataset_name>/.yaml` as it seems like these could be benchmarks that can be referenced to at least one paper?

prigoyal (approved):
looks really awesome!! we need to import it to phabricator, some inline comments + 2 additional things:
1. extend the documentation https://github.com/facebookresearch/vissl/blob/master/docs/source/vissl_modules/data.rst#using-builtin-datasets
2. documentation here https://github.com/facebookresearch/vissl/tree/master/vissl/data#expected-dataset-structure-for-imagenet-places205-places365

the docs could be a pain right now but it will help long term when we move to another vissl release in few months :) 

super awesome work @QuentinDuval <3

vissl/data/torchvision_dataset.py
QuentinDuval: It makes sense as a transform indeed ! I will rework that and get rid of the MNIST adapter which I did not find very satisfying either :)

QuentinDuval: This should be fixed as of 5a86a206e4f29d6c659db314cd55ee1e533c03c4. The code is better that way and can now be tested, which is really great :)

vissl/data/torchvision_dataset.py
QuentinDuval: I moved the names as constants in ef1f96f7d0f730ecc58cfb559f7894c38e14533f. This makes the code better and will allow us to have some documentation pointer. I am not sure the Enum would bring much in that case (the input field is a string, and should likely remain one since it is used for other formats of datasets as well, like "disk_folders").

QuentinDuval:
> 1. extend the documentation https://github.com/facebookresearch/vissl/blob/master/docs/source/vissl_modules/data.rst#using-builtin-datasets
> 2. documentation here https://github.com/facebookresearch/vissl/tree/master/vissl/data#expected-dataset-structure-for-imagenet-places205-places365

I added some documentation on the type of data sources as well as the format of the available torchvision datasets in f2803cec4653541b4a8ad417030c48e5194f5ef0.

vissl/data/torchvision_dataset.py
```diff
@@ -0,0 +1,88 @@
...
+            return CIFAR10(self.path, train=is_train_split)
+        elif self.dataset_name == TorchvisionDatasetName.CIFAR100:
+            return CIFAR100(self.path, train=is_train_split)
+        elif self.dataset_name == TorchvisionDatasetName.STL10:
+            stl_split = "train" if is_train_split else "test"
+            return STL10(self.path, split=stl_split)
+        elif self.dataset_name == TorchvisionDatasetName.MNIST:
+            return MNIST(root=self.path, train=is_train_split)
+        else:
+            raise ValueError(f"Unsupported dataset {self.dataset_name: str}")
```
prigoyal, line 60: nit: if we use the enum, we can remove the if/else need :) but it doesn't need to be addressed in this PR :)

vissl/data/torchvision_dataset.py
```diff
@@ -0,0 +1,88 @@
...
+    def __len__(self) -> int:
+        """
+        Size of the dataset
+        """
+        return self.num_samples()
+
+    def __getitem__(self, idx: int) -> Tuple[Image.Image, bool]:
+        """
+        Return the image at index 'idx' and whether the load was successful
+        Note: we do delayed loading of data to reduce the memory size
```
prigoyal, line 77: curious question: I didn't fully understand this piece. Are we saying dataloader will be slow?

vissl/data/ssl_transforms/img_pil_to_rgb_mode.py
```diff
@@ -0,0 +1,59 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
```
prigoyal, line 1: I think we need to make the transform name more specific otherwise users might mistakenly start using this transform for non-mnist datasets.

How about `img_pil_mnist_to_rgb_mode.py` ? Unless we think there will be more generic use of this transform, we can mention mnist specifically. Thoughts?

vissl/data/ssl_transforms/img_pil_to_rgb_mode.py
```diff
@@ -0,0 +1,59 @@
...
+
+from typing import Any, Dict, List, Union
+
+from PIL import Image
+from classy_vision.dataset.transforms import register_transform
+from classy_vision.dataset.transforms.classy_transform import ClassyTransform
+
+
+@register_transform("ImgPil2RGB")
+class ImgPil2RGB(ClassyTransform):
```
prigoyal, line 11: we should modify these name accordingly after we settle on the file name :)

prigoyal (commented):
The PR is looking really good. Just some inline small comments/questions. :) 

One more: We should format every single PR following the dev/linter.sh https://github.com/facebookresearch/vissl/blob/master/dev/README.md#option-1-use-devlintersh :) this will ensure that the linter in FB doesn't complain :)

configs/config/benchmark/linear_image_classification/datasets/cifar10.yaml
QuentinDuval: As discussed together, the best for now is to leave these configuration in <datasets> for as long as I am not able to get the exact results published in the papers (I am taking CLIP as reference here), so that the message is "we support those, but these are not part of the benchmark yet".

vissl/data/ssl_transforms/img_pil_to_rgb_mode.py
QuentinDuval: Yes, I agree with you, I will rename it accordingly :)

vissl/data/torchvision_dataset.py
QuentinDuval: No, it's just a bad copy-paste from "disk_dataset.py" :/ (I did not run into any performance issues)
I will remove this comment which is just not in line with the content of the function.

vissl/data/ssl_transforms/img_pil_to_rgb_mode.py
QuentinDuval: It should now be fixed in 1d0d4a1162346fde86938fa5f9e53291ef55f55e

vissl/data/ssl_transforms/img_pil_to_rgb_mode.py
QuentinDuval: It should now be fixed in 1d0d4a1162346fde86938fa5f9e53291ef55f55e

vissl/data/torchvision_dataset.py
QuentinDuval: It should now be fixed in 1d0d4a1162346fde86938fa5f9e53291ef55f55e

vissl/data/torchvision_dataset.py
QuentinDuval: The problem is that the `self.dataset_name` is a string and not an enumeration. And since it is plugged down as follow, I cannot really change the fact that the input is a string:

```
DATASET_SOURCE_MAP = {
    "disk_filelist": DiskImageDataset,
    "disk_folder": DiskImageDataset,
    "torchvision_dataset": TorchvisionDataset,
    "synthetic": SyntheticImageDataset,
}
```

And the string itself is given through the configuration, so it might be anything (and not necessarily a member of `TorchvisionDatasetName`). So in short, I think we need this default case and the associated input validation.

prigoyal:
looks great @QuentinDuval , let's reimport it to Phabricator and merge it! :)

prigoyal:
thank you @QuentinDuval , the PR is good to go. There are some linter/flake related issues on CI for this PR https://app.circleci.com/pipelines/github/facebookresearch/vissl/648/workflows/bd819aa0-08e6-46df-ad5e-4338534a037a/jobs/1283 and also take a look at https://www.internalfb.com/diff/D26453148 for `test_transforms.py` linter suggestions by fb linter. :) once those are resolved and CircleCI test is green, we can go ahead with landing it.

QuentinDuval:
> thank you @QuentinDuval , the PR is good to go. There are some linter/flake related issues on CI for this PR https://app.circleci.com/pipelines/github/facebookresearch/vissl/648/workflows/bd819aa0-08e6-46df-ad5e-4338534a037a/jobs/1283 and also take a look at [https://www.internalfb.com/diff/[D26453148](https://www.internalfb.com/intern/diff/D26453148)](https://www.internalfb.com/diff/D26453148) for `test_transforms.py` linter suggestions by fb linter. :) once those are resolved and CircleCI test is green, we can go ahead with landing it.

Excellent! Thank you @prigoyal :) I pushed a fix which should hopefully get rid of these formatting and linting errors: 655814f.

PR #182, opened 2021-02-09
Refactor distributed_launcher logic and simplify fb build

prigoyal:
Summary: we are bringing in submitit to VISSL (thanks to QuentinDuval ) and this required moving the `launch_distributed` to the core `vissl` library. Some refactorings needed to make the fblearner workflow adapt accordingly.

Differential Revision: D26340992


prigoyal:
will open a new one!

PR #183, opened 2021-02-09
Enable more frequent log flush to log.txt

prigoyal:
Summary:
:Facebook

When using manifold to write log.txt , the logs will buffer until 10MB in size and only then the training log is dumped. Or if the training finishes and buffer <10MB, the log will be dumped. The issue with this is that most of our training logs are like 2-5MB and we can't monitor an ongoing training if the logs are not flushed during the training.

It used to be the case that logs would flush with 1KB buffer but was changed to 10MB, for our trainings in vissl, 10KB is a good buffer size so for a 2KB training log, we will flush to manifold 200 times through training.

Reviewed By: sujitoc

Differential Revision: D26341684


prigoyal:
the code is already merged and some reason PR didn't close. closing it manually

PR #187, opened 2021-02-17
Update to isort 5.7.0

akainth015:
Our MLH team had issues committing files using the existing pre-commit hooks so I updated to `isort==5.7.0`, which no longer requires `seed-isort-config` and provides its own pre-commit hook. Once modified, the changes were run through `dev/linter.sh` which updated many of the older files that had been silently breaking the rules enforced for new files.

The most pertinent changes between `isort` 4 and 5 are listed below:

1. Only Python 3.6 is supported (this is okay, VISSL is 3.6 minimum anyways)
2. Command line options like `-sp` have been renamed to `--sp` to be less ambiguous
3. `isort` will not do anything by default. This has been remedied by specifying that it should check `"${DIR}"` in `dev/linter.sh`

After merging, contributors will need to reinstall the pre-commit hooks via `pre-commit install`.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -9,4 +9,5 @@ line_length = 88
 multi_line_output = 3
 use_parentheses = True
 lines_after_imports = 2
+known_third_party = PIL,bs4,classy_vision,cv2,detectron2,fairscale,faiss,fvcore,hydra,mock,nbconvert,nbformat,numpy,omegaconf,parameterized,pkg_resources,pycocotools,recommonmark,run_distributed_engines,scipy,setuptools,sklearn,sphinx_rtd_theme,tabulate,torch,torchvision,tqdm,utils
 skip_glob = build/*,third-party/*,.github/*
--- a/setup.cfg
+++ b/setup.cfg
@@ -9,5 +9,4 @@ line_length = 88
 multi_line_output = 3
 use_parentheses = True
 lines_after_imports = 2
-known_third_party = PIL,bs4,classy_vision,cv2,detectron2,faiss,fvcore,hydra,mock,nbconvert,nbformat,numpy,omegaconf,parameterized,pkg_resources,pycocotools,recommonmark,run_distributed_engines,scipy,setuptools,sklearn,sphinx_rtd_theme,tabulate,torch,torchvision,tqdm,utils
 skip_glob = build/*,third-party/*,.github/*
--- a/setup.cfg
+++ b/setup.cfg
@@ -9,5 +9,5 @@ line_length = 88
 multi_line_output = 3
 use_parentheses = True
 lines_after_imports = 2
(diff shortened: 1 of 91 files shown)


prigoyal:
thank you @anankan , can you click on "Import to Phabricator" ? we will need to run some more tests internally and ensure the isort change is aligned with the linters used in our fb development. Thank you so much for this PR! :)

setup.cfg
```diff
@@ -9,5 +9,4 @@ line_length = 88
 multi_line_output = 3
 use_parentheses = True
 lines_after_imports = 2
-known_third_party = PIL,bs4,classy_vision,cv2,detectron2,fairscale,faiss,fvcore,hydra,mock,nbconvert,nbformat,numpy,omegaconf,parameterized,pkg_resources,pycocotools,recommonmark,run_distributed_engines,scipy,setuptools,sklearn,sphinx_rtd_theme,tabulate,torch,torchvision,tqdm,utils
```
prigoyal, line 12: is this change enforced by isort 5.7?

tests/test_mlp.py
```diff
@@ -1,7 +1,8 @@
 import unittest
 
 import torch
-from vissl.models.heads import LinearEvalMLP, MLP
+
```
prigoyal, line 4: I am curious about the additional lines. Do you know isort is enforcing this?

setup.cfg
akainth015: I checked it out and it turns out it is not, I had just assumed (incorrectly) that since `seed-isort-config` was unnecessary that the option would be deprecated.

tests/test_mlp.py
akainth015: After your comment above, I figured it may have been influenced by my removing the `known_third_party` field. So I reverted the commit `22cf0d4` that applied the changes to other files and reintroduced the upgrade with the exception of removing `known_third_party`. These new lines were still applied to the original codebase, so it seems it is indeed something `isort` is enforcing. 

Since the history with the reverted commit would just reintroduce the changes, I have instead pushed a simple new commit to restore the original `setup.cfg`, since it appears to have no effect on the sorting. I now remember that I reasoned that since `known_third_party` was generated, and that since we were no longer generating anything to update it, that I would remove it from the codebase to avoid a stagnant line.

akainth015:
> thank you @anankan , can you click on "Import to Phabricator" ? we will need to run some more tests internally and ensure the isort change is aligned with the linters used in our fb development. Thank you so much for this PR! :)

I think you tagged the wrong person :stuck_out_tongue:, but I don't see an "Import to Phabricator" button anywhere on this pull request page.

prigoyal:
@akainth015 , let's open the PR in the MLH fork directly and we will look into it there.

Issue #198, opened 2021-02-26
Error in installation 


pclucas14:
Hi, 

I'm trying to install the package following the  instructions 

```
conda create -n vissl python=3.8
conda activate vissl
conda install -c pytorch pytorch=1.7.1 torchvision cudatoolkit=10.2
conda install -c vissl -c iopath -c conda-forge -c pytorch -c defaults apex vissl
```

However the last step seems to be failing : I'm getting 

```
(vissl) lucaspc@devfair0224:~/repos$ conda install -c vissl -c iopath -c conda-forge -c pytorch -c default
s apex vissl

Solving environment: failed

InvalidVersionSpecError: Invalid version spec: =2.7
```

I'm not sure where the error is coming from, there may be some version conflict.

Thanks, 
Lucas

pclucas14:
I also tried on another server, and it's still failing on the same step

```
(vissl) lpagec@dp-gpu1[pytorch]$conda install -c vissl -c iopath -c conda-forge -c pytorch -c defaults apex vissl
Collecting package metadata (current_repodata.json): done
Solving environment: failed with initial frozen solve. Retrying with flexible solve.
Solving environment: failed with repodata from current_repodata.json, will retry with next repodata source.
Collecting package metadata (repodata.json): done
Solving environment: failed with initial frozen solve. Retrying with flexible solve.
Solving environment: |
Found conflicts! Looking for incompatible packages.
This can take several minutes.  Press CTRL-C to abort.
Examining conflict for olefile ca-certificates typing_extensions mkl-service Examining conflict for olefile typing_extensions mkl-service numpy ninja six Examining conflict for olefile typing_extensions mkl-service numpy ninja six Examining conflict for olefile typing_extensions mkl-service numpy ninja six Examining conflict for olefile typing_extensions mkl-service numpy ninja six Examining conflict for olefile typing_extensions mkl-service numpy ninja six Examining conflict for olefile typing_extensions mkl-service numpy ninja six Examining conflict for olefile typing_extensions mkl-service numpy ninja six Examining conflict for olefile typing_extensions mkl-service numpy ninja six Examining conflict for olefile typing_extensions mkl-service numpy freetype nExamining conflict forExamining conflict for olefile typing_extensions mkl-service numpy freetype ninja six libstdcxx-ng nump-
```

962086838:
Same issue on my Ubuntu 18.04 server. Here's my log. 

[...]

PR #208, opened 2021-03-01
dependencies for conda package testing

bottler:
Now opencv and tensorboard are not official dependencies, make sure they are available for the conda tests. This is part of the followup to #198 (conda not installing nicely), specifically PRs #202 and #205 which removed them as deps.


bottler:
(I have put the new packages on anaconda and removed the old ones.)

PR #214, opened 2021-03-04
[fix] configuration for low_shot experiments for which the transformation pipeline has 2 consecutive resize

QuentinDuval:
This is a rather simplification: two resize consecutive to each other can be merged into one single.

I tested that there are no side effects from that: the two pipelines (before and after simplification) lead to the same output vector.


prigoyal:
looks great! let's import to Phabricator and merge it :)

prigoyal (approved):
thank you @QuentinDuval :) let's import to phabricator and merge!!!

PR #215, opened 2021-03-04
[refactor] re-organizing the fulltune configurations under a common folder

QuentinDuval:
Full tuning is a very useful benchmark benchmark. The number of datasets to be tested in full tune mode will likely grow, and it would be great to have the current full tuning configuration grouped under a single folder (much like the "linear_classification" benchmarks).

This is a simple move of files, but which might be hard to integrate due to potential conflicts (which means that if it must be integrated, we must integrate it right after a rebase)


prigoyal (approved):
looks great! let's import to phabricator and merge it!!

also , we should grep once if there are other references to the old folders and update them! 

thank you @QuentinDuval :)

QuentinDuval:
> looks great! let's import to phabricator and merge it!!
> 
> also , we should grep once if there are other references to the old folders and update them!
> 
> thank you @QuentinDuval :)

The references in the documentation should be in 5d9c7e3. I hope I did not miss some !

QuentinDuval:
Closed (rebased at https://github.com/facebookresearch/vissl/pull/226)

PR #216, opened 2021-03-04
[Proposal] Improve the DATA_LIMIT atribute to be able to handle more uses cases

QuentinDuval:
This PR is a draft, pushed for visibility and discussion.

The additional uses cases I propose to support are:
- being able to sub-select part of a dataset in a balanced way (each label is included the same number of time)
- being able to sub-select exclusive parts of the same dataset (for instance to have a validation set that does not intersect with a training set, useful for HP searches)
- make sure that this sub-sampling is deterministic (same seed across all distributed workers)

This would avoid having to create sub-sets of datasets such as ImageNet to test on 1% of each label for instance. It would also allow to benchmark SSL algorithms on low data regime in a more flexible way.

/!\ This PR introduces a breaking change (DATA_LIMIT is not an integer anymore but a structure)

This PR includes:
- unit tests for the sub-sampling strategies
- update of all configuration using the DATA_LIMIT attribute

### DIFF
--- a/vissl/config/defaults.yaml
+++ b/vissl/config/defaults.yaml
@@ -250,8 +250,26 @@ config:
       # accomodate the smoothed labels. See
       # LOSS.cross_entropy_multiple_output_single_target for more information.
       #
-      # limit the amount of data used in training. If set to -1, full dataset is used.
+      # Limit the amount of data used in training by specifying a range of type "start + number of elements":
+      #   DATA_LIMIT: N      -1 means that the range will extend to the end, otherwise sample N elements
+      #   DATA_LIMIT_SKIP: S  0 means that the range will start from the Sth element, effectively removing S element
+      #   DATA_LIMIT_SAMPLING:
+      #     SEED: 0           specifies the seed used to sample these elements from the full dataset
+      #     BALANCED: False   whether or not each class should appear equally in the sample (requires labels)
+      #
+      # Example: to select a range of 500 samples for validation, skipping the first 1000 samples (say these are
+      # already used in the training split) and sub-sampling these elements such that each class appears equally:
+      # DATA_LIMIT: 500
+      # DATA_LIMIT_SKIP: 1000
+      # DATA_LIMIT_SAMPLING:
+      #   SEED: 0
+      #   BALANCED: True
       DATA_LIMIT: -1
+      DATA_LIMIT_SKIP: 0
+      DATA_LIMIT_SAMPLING:
+        SEED: 0
+        BALANCED: False
+
       # whether the data specified (whether file list or directory) should be copied locally
       # on the machine where training is happening.
       COPY_TO_LOCAL_DISK: False
@@ -286,6 +304,10 @@ config:
       COLLATE_FUNCTION: "default_collate"
       COLLATE_FUNCTION_PARAMS: {}
       DATA_LIMIT: -1
+      DATA_LIMIT_SKIP: 0
+      DATA_LIMIT_SAMPLING:
+        SEED: 0
+        BALANCED: False
       DATASET_NAMES: ["imagenet1k_folder"]
       COPY_TO_LOCAL_DISK: False
       COPY_DESTINATION_DIR: ""
--- a/vissl/config/defaults.yaml
+++ b/vissl/config/defaults.yaml
@@ -250,25 +250,24 @@ config:
       # accomodate the smoothed labels. See
       # LOSS.cross_entropy_multiple_output_single_target for more information.
       #
-      # Limit the amount of data used in training by specifying a range of type "start + number of elements":
-      #   DATA_LIMIT: N      -1 means that the range will extend to the end, otherwise sample N elements
-      #   DATA_LIMIT_SKIP: S  0 means that the range will start from the Sth element, effectively removing S element
-      #   DATA_LIMIT_SAMPLING:
-      #     SEED: 0           specifies the seed used to sample these elements from the full dataset
-      #     BALANCED: False   whether or not each class should appear equally in the sample (requires labels)
+      # Limit the amount of data used in training. If set to -1, full dataset is used.
+      #
+      DATA_LIMIT: -1
+      #
+      # Specifies how the DATA_LIMIT samples are sampled
       #
       # Example: to select a range of 500 samples for validation, skipping the first 1000 samples (say these are
       # already used in the training split) and sub-sampling these elements such that each class appears equally:
       # DATA_LIMIT: 500
-      # DATA_LIMIT_SKIP: 1000
       # DATA_LIMIT_SAMPLING:
       #   SEED: 0
       #   BALANCED: True
-      DATA_LIMIT: -1
-      DATA_LIMIT_SKIP: 0
+      #   SKIP: 1000
+      #
       DATA_LIMIT_SAMPLING:
         SEED: 0
         BALANCED: False
+        SKIP: 0
 
       # whether the data specified (whether file list or directory) should be copied locally
       # on the machine where training is happening.
@@ -304,10 +303,10 @@ config:
       COLLATE_FUNCTION: "default_collate"
       COLLATE_FUNCTION_PARAMS: {}
       DATA_LIMIT: -1
-      DATA_LIMIT_SKIP: 0
       DATA_LIMIT_SAMPLING:
         SEED: 0
         BALANCED: False
+        SKIP: 0
       DATASET_NAMES: ["imagenet1k_folder"]
       COPY_TO_LOCAL_DISK: False
       COPY_DESTINATION_DIR: ""
--- a/vissl/config/defaults.yaml
+++ b/vissl/config/defaults.yaml
@@ -261,12 +261,12 @@ config:
       # DATA_LIMIT: 500
       # DATA_LIMIT_SAMPLING:
       #   SEED: 0
-      #   BALANCED: True
-      #   SKIP: 1000
+      #   IS_BALANCED: True
+      #   SKIP_NUM_SAMPLES: 1000
       #
       DATA_LIMIT_SAMPLING:
         SEED: 0
-        BALANCED: False
+        IS_BALANCED: False
         SKIP_NUM_SAMPLES: 0
 
       # whether the data specified (whether file list or directory) should be copied locally
@@ -305,8 +305,8 @@ config:
       DATA_LIMIT: -1
       DATA_LIMIT_SAMPLING:
         SEED: 0
-        BALANCED: False
-        SKIP: 0
+        IS_BALANCED: False
+        SKIP_NUM_SAMPLES: 0
       DATASET_NAMES: ["imagenet1k_folder"]
       COPY_TO_LOCAL_DISK: False
       COPY_DESTINATION_DIR: ""
--- a/vissl/config/defaults.yaml
+++ b/vissl/config/defaults.yaml
@@ -267,7 +267,7 @@ config:
       DATA_LIMIT_SAMPLING:
         SEED: 0
         BALANCED: False
-        SKIP: 0
+        SKIP_NUM_SAMPLES: 0
 
       # whether the data specified (whether file list or directory) should be copied locally
       # on the machine where training is happening.
--- a/vissl/data/data_helper.py
+++ b/vissl/data/data_helper.py
@@ -36,17 +36,23 @@ def with_temporary_numpy_seed(seed: int):
     np.random.set_state(random_state)
 
 
-def unbalanced_sub_sampling(total_size: int, nb_samples: int, skip_samples: int = 0, seed: int = 0) -> np.ndarray:
+def unbalanced_sub_sampling(
+    total_size: int, nb_samples: int, skip_samples: int = 0, seed: int = 0
+) -> np.ndarray:
     """
     Given an original dataset of size 'total_size', sub_sample part of the dataset such that
     the sub sampling is deterministic (identical across distributed workers)
     Return the selected indices
     """
     with with_temporary_numpy_seed(seed):
-        return np.random.choice(total_size, size=skip_samples+nb_samples, replace=False)[skip_samples:]
+        return np.random.choice(
+            total_size, size=skip_samples + nb_samples, replace=False
+        )[skip_samples:]
 
 
-def balanced_sub_sampling(labels: np.ndarray, nb_samples: int, skip_samples: int = 0, seed: int = 0) -> np.ndarray:
+def balanced_sub_sampling(
+    labels: np.ndarray, nb_samples: int, skip_samples: int = 0, seed: int = 0
+) -> np.ndarray:
     """
     Given some labels, sub_sample a part of the labels such that:
     - the number of samples of each label differs by at most one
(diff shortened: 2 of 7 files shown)


prigoyal (approved):
super cool and makes a lot of sense IMO!! thank you @QuentinDuval :)

one small nit: `NB_SAMPLES` -> `NUM_SAMPLES` . We use `num` everywhere in VISSL. consistency enforcement!

there are some bash scripts and possibly docs that might require update with this PR. can you grep once to double check? I love this change :)

QuentinDuval:
> super cool and makes a lot of sense IMO!! thank you @QuentinDuval :)
> 
> one small nit: `NB_SAMPLES` -> `NUM_SAMPLES` . We use `num` everywhere in VISSL. consistency enforcement!
> 
> there are some bash scripts and possibly docs that might require update with this PR. can you grep once to double check? I love this change :)

Makes sense :) I will rename the attribute and look inside the docs (I searched the code, but I forgot to search the doc so there is likely some change to be made)

QuentinDuval:
> super cool and makes a lot of sense IMO!! thank you @QuentinDuval :)
> 
> one small nit: `NB_SAMPLES` -> `NUM_SAMPLES` . We use `num` everywhere in VISSL. consistency enforcement!
> 
> there are some bash scripts and possibly docs that might require update with this PR. can you grep once to double check? I love this change :)

Renamed to NUM_SAMPLES, and corrected the references, following d394af5 :)

prigoyal:
hmm the circleCI gpu error triggered is very interesting. Let's investigate that. We can also try re-running the test in case it was a flaky instance.

QuentinDuval:
> hmm the circleCI gpu error triggered is very interesting. Let's investigate that. We can also try re-running the test in case it was a flaky instance.

I will look into this, it might have broken some configurations.

vissl/data/data_helper.py
```diff
@@ -24,6 +24,65 @@ def get_mean_image(crop_size):
     return img
 
 
+@contextlib.contextmanager
+def with_temporary_numpy_seed(seed: int):
+    """
+    Context manager to run a specific portion of code with a given seed:
+    resumes the numpy state after the execution of the block
+    """
+    random_state = np.random.get_state()
```
prigoyal, line 33: rename to `orig_random_state` :)

vissl/data/data_helper.py
```diff
@@ -24,6 +24,65 @@ def get_mean_image(crop_size):
     return img
 
 
+@contextlib.contextmanager
+def with_temporary_numpy_seed(seed: int):
```
prigoyal, line 28: nit: rename `sampling_seed` for clarity that it's not the other training seed :)

vissl/data/data_helper.py
```diff
@@ -24,6 +24,65 @@ def get_mean_image(crop_size):
...
+    resumes the numpy state after the execution of the block
+    """
+    random_state = np.random.get_state()
+    np.random.seed(seed)
+    yield
+    np.random.set_state(random_state)
+
+
+def unbalanced_sub_sampling(
+    total_size: int, nb_samples: int, skip_samples: int = 0, seed: int = 0
```
prigoyal, line 40: nit: `total_size` -> `total_num_samples`

vissl/data/data_helper.py
```diff
@@ -24,6 +24,65 @@ def get_mean_image(crop_size):
...
+    - the sub sampling is deterministic (identical across distributed workers)
+    Return the selected indices
+    """
+    groups = {}
+    for i, label in enumerate(labels):
+        groups.setdefault(label, []).append(i)
+
+    unique_labels = sorted(groups.keys())
+    sq, sr = divmod(skip_samples, len(unique_labels))
+    q, r = divmod(nb_samples, len(unique_labels))
```
prigoyal, line 68: is it feasible to make the `sq, sr, q, r` more expanded names? :)

vissl/data/data_helper.py
```diff
@@ -24,6 +24,65 @@ def get_mean_image(crop_size):
...
+    labels: np.ndarray, nb_samples: int, skip_samples: int = 0, seed: int = 0
+) -> np.ndarray:
+    """
+    Given some labels, sub_sample a part of the labels such that:
+    - the number of samples of each label differs by at most one
+    - the sub sampling is deterministic (identical across distributed workers)
+    Return the selected indices
+    """
+    groups = {}
+    for i, label in enumerate(labels):
```
prigoyal, line 63: does this have any assumption on the labels type being list, numpy array etc ?

vissl/data/ssl_dataset.py
```diff
@@ -241,11 +265,17 @@ def __getitem__(self, idx):
             self._load_labels()
             self._labels_init = True
 
+        if self.data_limit >= 0:
+            self._init_subset()
```
prigoyal, line 269: can we add the check `if not self._subset_initialized`: right here? so that it clear that we only call the init_subset if needed ? :)

vissl/data/ssl_dataset.py
```diff
@@ -241,11 +265,17 @@ def __getitem__(self, idx):
...
+            subset_idx = self.sub_set[idx]
+        else:
+            subset_idx = idx
```
prigoyal, line 272: we could write this whole as
```
item_index = idx
if self.data_limit >= 0:
    if not self._subset_initialized:
         self._init_subset()
    item_index = self.subset[idx]
```

what do you think?

vissl/data/disk_dataset.py
```diff
@@ -87,13 +87,6 @@ def _load_data(self, path):
             # Avoid creating it over and over again.
             self.is_initialized = True
 
-        if self.cfg["DATA"][self.split]["DATA_LIMIT"] > 0:
```
prigoyal, line 90: I am a little bit worried about moving this out to the ssl_dataset.py. The reason is because we have some internal data sources that don't apply the data limit in similar way and could break training for them. Since our intent is to support the data limit like this for the synthetic and disk_dataset, can we put the `init_subset()` in data helper and call the function from respective disk_dataset.py and `synthetic_dataset.py` ? any other thoughts/suggestions. Sorry if I am missing some piece :)

prigoyal (commented):
Thank you @QuentinDuval , looks really good and a very  nice proposal too :)

I left some inline thoughts . lmk what you think :) we can aim to land this soon within a day.

vissl/data/synthetic_dataset.py
```diff
@@ -19,15 +19,16 @@ class SyntheticImageDataset(Dataset):
         data_source (string, Optional): data source ("synthetic") [not used]
     """
 
-    def __init__(self, cfg, path, split, dataset_name, data_source="synthetic"):
+    DEFAULT_SIZE = 50_000
```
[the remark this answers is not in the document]
QuentinDuval: I thought 50_000 would be more readable than 50000.

I checked and the associated PEP that introduces this feature in Python (https://www.python.org/dev/peps/pep-0515/) has been integrated in Python 3.6 (so it should work with all VISSL supported Python versions).

vissl/data/data_helper.py
QuentinDuval: Indeed... sorry for that. I wanted to do it, but it slipped my mind and forgot to do it. I will fix this !

vissl/data/data_helper.py
QuentinDuval: The type hints says that the "labels" should be a numpy array, but this could be relaxed to any kind of Sequence. The other assumption is that the labels inside it should be hashable.

vissl/data/disk_dataset.py
QuentinDuval: Do you have some pointers to these internal data sources ?

I mostly moved the code to `ssl_dataset.py` for two reasons:
- I saw that everything was scheduled in `ssl_dataset` (but as you point out, I clearly missing something) and found it would be a good way to factorise the code and the feature (we cannot random sample twice)
- The balanced random sampling (same amount of sample by classes) requires to have access to the labels, and these labels are only available in ssl_dataset

vissl/data/ssl_dataset.py
```diff
@@ -60,20 +62,24 @@ class GenericSSLDataset(Dataset):
...
         self.data_paths = []
         self.label_paths = []
+        self.sub_set = None
         self.batchsize_per_replica = self.cfg["DATA"][split]["BATCHSIZE_PER_REPLICA"]
         self.data_sources = self.cfg["DATA"][split].DATA_SOURCES
         self.label_sources = self.cfg["DATA"][split].LABEL_SOURCES
         self.dataset_names = self.cfg["DATA"][split].DATASET_NAMES
         self.label_type = self.cfg["DATA"][split].LABEL_TYPE
+        self.data_limit = self.cfg["DATA"][split].DATA_LIMIT
+        self.data_limit_sampling = self.cfg["DATA"][split].DATA_LIMIT_SAMPLING
```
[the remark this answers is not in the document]
QuentinDuval: Shouldn't the "default.yaml" make sure that there will always be such a node?

In any case, I can indeed use ".get()", in which case the best is probably to return a default data-structure, with the options such as "not balanced" and "seed = 0" and "skip = 0". What do you think?

vissl/data/ssl_dataset.py
```diff
@@ -60,20 +62,24 @@ class GenericSSLDataset(Dataset):
...
 
-    def __init__(self, cfg, split, dataset_source_map):
+    def __init__(self, cfg: AttrDict, split: str, dataset_source_map):
         self.split = split
         self.cfg = cfg
         self.data_objs = []
         self.label_objs = []
         self.data_paths = []
         self.label_paths = []
+        self.sub_set = None
```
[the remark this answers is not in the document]
QuentinDuval: Oh, it will affect both the images and the labels: it effectively reduces the size of the dataset by mapping the range [0, K) to [0, N), where K < N is the DATA_LIMIT value.

vissl/data/ssl_dataset.py
```diff
@@ -221,7 +227,25 @@ def _load_labels(self):
                 raise ValueError(f"unknown label source: {label_source}")
             self.label_objs.append(labels)
 
-    def __getitem__(self, idx):
+    def _init_subset(self):
+        if not self._subset_initialized:
+            if not self.data_limit_sampling.BALANCED:
+                self.sub_set = unbalanced_sub_sampling(
+                    total_size=len(self.data_objs[0]),
```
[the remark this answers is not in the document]
QuentinDuval: For the moment, there are a few that seems to be made with the assumption that we have 1 data source, or that all data sources have the same length. I think this code does not make further assumption (*).

I am actually not quite sure where to push this in the future. The way I see it, we need to rework a bit the way we combine datasets. In particular, combining several datasets (of different length) means reworking the `ssl_dataset.py` module such that:
- combining means concatenating (which is different from the current behaviour, which, if we have 2 dataset, is such that we draw two samples, one for each dataset, at each call to __getitem__)
- labels of different kind can be concatenated (meaning that 2 datasets with different labels means twice as much labels, and optionally deal with the overlap?)

(*) In the current state of the code, the `_init_subset` function does not actually requires to have just one dataset: we can have several, as long as they have the same length, but the sampling will be performed on the first one (which means that the "balanced" constraint might not be respected for the second source, but overall the code will run fine).

What we can do is add an assert for the balanced in case we have either "no label" or "more than 1 data source".

vissl/config/defaults.yaml
```diff
@@ -250,8 +250,25 @@ config:
...
+      # DATA_LIMIT: 500
+      # DATA_LIMIT_SAMPLING:
+      #   SEED: 0
+      #   BALANCED: True
+      #   SKIP: 1000
+      #
+      DATA_LIMIT_SAMPLING:
+        SEED: 0
+        BALANCED: False
+        SKIP: 0
```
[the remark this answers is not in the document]
QuentinDuval: 100% agree, the current name is misleading, I will rename this :)

vissl/config/defaults.yaml
```diff
@@ -250,8 +250,25 @@ config:
...
+      # already used in the training split) and sub-sampling these elements such that each class appears equally:
+      # DATA_LIMIT: 500
+      # DATA_LIMIT_SAMPLING:
+      #   SEED: 0
+      #   BALANCED: True
+      #   SKIP: 1000
+      #
+      DATA_LIMIT_SAMPLING:
+        SEED: 0
+        BALANCED: False
```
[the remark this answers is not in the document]
QuentinDuval: I think `IS_BALANCED` should suffices as the `SAMPLING` part is already implied by the father node. So if a user wants to use the override syntax, the command line will already contain it:

`config.DATA.TRAIN.DATA_LIMIT_SAMPLING.IS_BALANCED=True`

What do you think?

vissl/data/data_helper.py
QuentinDuval: It will be done :)

vissl/data/synthetic_dataset.py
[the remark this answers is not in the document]
prigoyal: oh nice! this is amazing and I learn something! thank you Quentin :)

vissl/data/ssl_dataset.py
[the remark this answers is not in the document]
prigoyal: yep that sounds good. And I also agree that `default.yaml` should take care of it. However, I discovered over the last few days that some projects are using VISSL and somehow their code breaks if we don't use .get(). It's unclear that if they use new vissl code they why their defaults.yaml is not getting updated. Something to dig into later separately for me.

vissl/data/data_helper.py
prigoyal: docstring is perfect :) also , since labels can be many things, we can remove the type hint for labels or use the `Union` to compose the typehint :)

vissl/data/disk_dataset.py
prigoyal: yes the suggestion was to remove it move it to synthetic_dataset and disk_dataset. Let me share some examples . 
1) Essentially https://www.internalfb.com/intern/diffusion/FBS/browsefile/master/fbcode/deeplearning/projects/ssl_framework/vissl/fb/data/fairstore_dataset.py?commit=beaa5fafc669889804877ed3efd62c9df87e1507&lines=103-104 doesn't have the concept of DATA_LIMIT 
2) an internal change in https://www.internalfb.com/intern/diffusion/FBS/browsefile/master/fbcode/deeplearning/projects/ssl_framework/vissl/fb/data/everstore_dataset.py?commit=beaa5fafc669889804877ed3efd62c9df87e1507&lines=162-165 will be needed (a diff from fbcode) to adapt to the new call :)

vissl/config/defaults.yaml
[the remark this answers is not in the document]
prigoyal: sounds good :)

prigoyal:
looks great and good to go! just 2 open comments: 1) use of `.get()` 2) renaming the new config variables (IS_BALANCED) :)

QuentinDuval:
> looks great and good to go! just 2 open comments: 1) use of `.get()` 2) renaming the new config variables (IS_BALANCED) :)

Indeed, I really need to fix these things and I will see to it Monday :-) Thank you @prigoyal for your review !

QuentinDuval:
> > looks great and good to go! just 2 open comments: 1) use of `.get()` 2) renaming the new config variables (IS_BALANCED) :)
> 
> Indeed, I really need to fix these things and I will see to it Monday :-) Thank you @prigoyal for your review !

Both issues should be fixed at 6c4788e :)

vissl/data/ssl_dataset.py
```diff
@@ -221,7 +234,53 @@ def _load_labels(self):
...
+        This function will create a mapping from [0, K) to [0, N), using the
+        parameters specified in the DATA_LIMIT_SAMPLING configuration. This
+        mapping is then cached and used for all __getitem__ calls to map
+        the external indices from [0, K) to the internal [0, N) indices.
+
+        This function makes the assumption that there is one data source only
+        or that all data sources have the same length (same as __getitem__).
+        """
+
+        valid_datasets = {
```
prigoyal, line 250: I think we should move this check down to line 305 . We should just not call the init_subset function if the conditions are not met.

prigoyal (commented):
one small comment inline and good to go otherwise! :) :)

vissl/data/ssl_dataset.py
QuentinDuval: It should be fixed at 5586fac !

prigoyal:
the fvcore issue has struck here too. Can you rebase on master and push again? A recent landed PR will help with the circleCI tests.

PR #224, opened 2021-03-09
Installation instructions for VISSL in source mode (dev should be surrounded by quotes)

QuentinDuval:


INSTALL.md
```diff
@@ -122,7 +122,7 @@ pip install opencv-python
 pip uninstall -y classy_vision
 pip install classy-vision@https://github.com/facebookresearch/ClassyVision/tarball/master
 # install vissl dev mode (e stands for editable)
-pip install -e .[dev]
+pip install -e ".[dev]"
```
prigoyal, line 125: nice !

for curiosity: when does it break if we don't use the quotes ?

prigoyal (commented):
looks great, small inline curious comment :)

let's also update the docs/installation.rst file to reflect this change.

INSTALL.md
QuentinDuval: > nice !
> 
> for curiosity: when does it break if we don't use the quotes ?

I am using ZSH, and I get this error when trying to run the command:

```
$ pip install -e .[dev]
zsh: no matches found: .[dev]
```

Surrounding it with quote makes the target not interpreted, otherwise it seems that zsh will try to interpret it as a regex.

QuentinDuval:
> looks great, small inline curious comment :)
> 
> let's also update the docs/installation.rst file to reflect this change.

Indeed, I keep forgetting the docs !

It should be fixed following f08d2fa.

prigoyal (approved):

PR #230, opened 2021-03-10
Barlow Twins implementation

OlivierDehaene:
## Required (TBC)

- [X] BarlowTwinsLoss and Criterion
- [X] Loss unit and integration tests
- [ ] Documentation
- [ ] Default configs
- [ ] Benchmarks

## Optional

- [ ] Support multi crop


configs/config/test/integration_test/quick_barlow_twins.yaml
```diff
@@ -0,0 +1,121 @@
...
+      weight_decay: 0.0000015
+      momentum: 0.9
+      nesterov: False
+      num_epochs: 2
+      regularize_bn: False
+      regularize_bias: True
+      param_schedulers:
+        lr:
+          auto_lr_scaling:
+            auto_scale: true # TODO: not sure about this value
```
OlivierDehaene, line 94: @prigoyal If we are using schedulers, can we still use the `auto_lr_scaling`?

The `linear` scheduler end value and `cosine` scheduler start and end value will need to be re-scaled too. This would be impossible to do automatically, wouldn't it?

OlivierDehaene: You can ignore my comment. I found what I was looking for in [hydra_config.py#get_scaled_lr_scheduler](https://github.com/facebookresearch/vissl/blob/4cffc1c37b236371e4459278c0d7997bfdcd7e8d/vissl/utils/hydra_config.py#L190).

configs/config/test/integration_test/quick_barlow_twins.yaml
```diff
@@ -0,0 +1,121 @@
...
+          # TODO: not sure about these values
+          schedulers:
+            - name: linear
+              start_value: 0.0
+              end_value: 1.6
+            - name: cosine
+              start_value: 1.6
+              end_value: 0.002
+          update_interval: step
+          interval_scaling: [rescaled, fixed]
```
OlivierDehaene, line 107: @prigoyal I could not find where `interval_scaling` is used in the code.

prigoyal: @OlivierDehaene , the `interval_scaling` decides how to use the schedulers in a composite setting. It seems like a recent refactor of `param_schedulers` in classy vision moved the documentation here https://github.com/facebookresearch/fvcore/blob/master/fvcore/common/param_scheduler.py#L295-L324

OlivierDehaene: Thank you!

OlivierDehaene:
@prigoyal, @jzbontar,

Using the [following configuration](https://github.com/facebookresearch/vissl/pull/230/files#diff-0b4ecea4c3a718b39b18b9d428fc22e914487d7022cffc20efddccf1e2f7fb55) for pretraining, and [this one](https://github.com/facebookresearch/vissl/blob/master/configs/config/debugging/benchmark/linear_image_classification/eval_resnet_8gpu_transfer_imagenette_160.yaml) for evaluation, I obtain **85.7 Top 1 Accuracy** on Imagenette 160. 
Is it in the ballpark of what you would expect on this dataset? 

If so, I think [the BarlowTwinsLoss and Criterion](https://github.com/facebookresearch/vissl/pull/230/files#diff-1f4ffba2bdf80d61bafb8230356f67413df274584b94e17ca285ec1d7aa3dc93) are ready for review!

QuentinDuval:
> @prigoyal, @jzbontar,
> 
> Using the [following configuration](https://github.com/facebookresearch/vissl/pull/230/files#diff-0b4ecea4c3a718b39b18b9d428fc22e914487d7022cffc20efddccf1e2f7fb55) for pretraining, and [this one](https://github.com/facebookresearch/vissl/blob/master/configs/config/debugging/benchmark/linear_image_classification/eval_resnet_8gpu_transfer_imagenette_160.yaml) for evaluation, I obtain **85.7 Top 1 Accuracy** on Imagenette 160.
> Is it in the ballpark of what you would expect on this dataset?
> 
> If so, I think [the BarlowTwinsLoss and Criterion](https://github.com/facebookresearch/vissl/pull/230/files#diff-1f4ffba2bdf80d61bafb8230356f67413df274584b94e17ca285ec1d7aa3dc93) are ready for review!

I do not know for Barlow Twins specifically (@jzbontar can answer on that more precisely), but this is definitely in the same ball park than my experiments with SimCLR pre-trained and then evaluated on Imagenette 160 (where the typical results I got are between 85% to 88%).

vissl/losses/barlow_twins_loss.py
```diff
@@ -0,0 +1,124 @@
...
+
+@register_loss("barlow_twins_loss")
+class BarlowTwinsLoss(ClassyLoss):
+    """
+    This is the loss which was proposed in the Barlow Twins https://arxiv.org/abs/2103.03230v1
+    paper.
+    See the paper for the details on the loss.
+
+    Config params:
+        lambda_ (float): weight on the off-diagonal terms. It controls the trade-off between
```
prigoyal, line 20: if this is config parameter, do we require `_` in `lambda_`  ? or can we rename it `lambda` ?

vissl/losses/barlow_twins_loss.py
```diff
@@ -0,0 +1,124 @@
...
+        return pprint.pformat(repr_dict, indent=2)
+
+
+class BarlowTwinsCriterion(nn.Module):
+    """
+    The criterion corresponding to the Barlow Twins loss as defined in the paper
+    https://arxiv.org/abs/2103.03230v1.
+
+    Args:
+        lambda_ (float): weight on the off-diagonal terms. It controls the trade-off between
```
prigoyal, line 65: same comment as above :)

vissl/losses/barlow_twins_loss.py
```diff
@@ -0,0 +1,124 @@
...
+        embedding_dim (int): dimensionality of the representation
+    """
+
+    def __init__(self, lambda_: float, embedding_dim: int):
+        super(BarlowTwinsCriterion, self).__init__()
+
+        self.use_gpu = get_cuda_device_index() > -1
+        self.lambda_ = lambda_
+        self.embedding_dim = embedding_dim
+        self.num_copies = 2
```
prigoyal, line 76: can we make this a parameter from config file?  we can set the default value to 2. Unless there are reservations that the loss wouldn't work for >2 copies in which case, hardcoding is fine.

vissl/losses/barlow_twins_loss.py
```diff
@@ -0,0 +1,124 @@
...
+
+        identity_matrix = torch.eye(embedding_dim)
```
prigoyal, line 78: can we make this `self.identity_matrix` instead of on line 82 ?

vissl/losses/barlow_twins_loss.py
```diff
@@ -0,0 +1,124 @@
...
+        if self.use_gpu:
+            identity_matrix = identity_matrix.cuda(non_blocking=True)
+
+        self.identity_matrix = identity_matrix
+
+    def forward(self, embedding: torch.Tensor):
+        """
+        Calculate the loss. Operates on embeddings tensor.
```
prigoyal, line 86: Can you add a docstring on the expected `embedding` tensor shape? for the docs if one doesn't consult the code :)

vissl/losses/barlow_twins_loss.py
```diff
@@ -0,0 +1,124 @@
...
+        """
+        assert embedding.ndim == 2
```
prigoyal, line 88: let's throw an assert error message 
`assert embedding.ndim == 2, f"Incorrect embedding dimension: {embedding.ndim} but expected 2"`

vissl/losses/barlow_twins_loss.py
```diff
@@ -0,0 +1,124 @@
...
+        assert embedding.shape[1] == int(self.embedding_dim)
```
prigoyal, line 89: let's add an error message similar to above. This will help facilitate debugging :)

vissl/losses/barlow_twins_loss.py
```diff
@@ -0,0 +1,124 @@
...
+
+        batch_size = embedding.shape[0]
+        L = self.lambda_
+        num_copies = self.num_copies
```
prigoyal, line 93: do we need this or can we directly use the `self.lambda_` and `self.num_copies`?

vissl/losses/barlow_twins_loss.py
```diff
@@ -0,0 +1,124 @@
...
+        assert batch_size % num_copies == 0, "Batch size should be divisible by num_copies"
```
prigoyal, line 94: can we expand the assert message a bit more to specify what `batch_size` value was?

vissl/losses/barlow_twins_loss.py
```diff
@@ -0,0 +1,124 @@
...
+
+        # normalize embeddings along the batch dimension
+        embedding_normed = (embedding - embedding.mean(dim=0)) / (embedding.std(dim=0) + 1e-16)
```
prigoyal, line 97: is it possible to capture the `1e-16` as `self.epsilon` ?

vissl/losses/barlow_twins_loss.py
```diff
@@ -0,0 +1,124 @@
...
+
+        Args:
+            loss_config: configuration for the loss
+
+        Returns:
+            BarlowTwinsLoss instance.
+        """
+        return cls(loss_config)
+
+    def forward(self, output, target):
```
prigoyal, line 46: can we add type hints for `output` and `target`? for example https://github.com/facebookresearch/vissl/blob/master/vissl/losses/swav_loss.py#L83

prigoyal (commented):
thank you so much @OlivierDehaene  this is looking great. I have left some inline comments. Please take a look at them :) Also a next step I see:
Can you follow the https://github.com/facebookresearch/vissl/blob/master/.github/CONTRIBUTING.md#coding-style to try the dev/linter.sh on this PR ?

Afterwards, I think the open items in the PR checklist are next steps. We should aim to provide a config file for at least ImageNet. Further, a model in the model_zoo would be great .

vissl/losses/barlow_twins_loss.py
OlivierDehaene: Yes as `lambda` is a reserved name in Python.

vissl/losses/barlow_twins_loss.py
OlivierDehaene: I think the loss doesn't really make sense for values > 2 but I didn't really think about it. At least, we cannot compare to the identity matrix if num_copies > 2 so I believe it should be handled in another class if it is indeed possible like it is done for the SimCLR loss.

vissl/losses/barlow_twins_loss.py
OlivierDehaene: Ok. I will update this line and the one in the SimCLR criterion.

vissl/losses/barlow_twins_loss.py
OlivierDehaene: Same as above.

vissl/losses/barlow_twins_loss.py
OlivierDehaene: I used the same syntax as in the SimCLR loss. I will update with the syntax you propose.

OlivierDehaene:
@prigoyal,

My implementation is based and consistent with the SimCLR loss implementation. A lot of the comments you added are relevant for both ([for example the style comments](https://github.com/facebookresearch/vissl/blob/master/vissl/losses/simclr_info_nce_loss.py#L127)). Do you want me to patch both at the same time?

vissl/losses/barlow_twins_loss.py
OlivierDehaene: I also think `lambda_` is ugly but I didn't find a better alternative =(

prigoyal:
> @prigoyal,
> 
> My implementation is based on the SimCLR loss implementation. A lot of the comments you added are relevant for both ([for example the style comments](https://github.com/facebookresearch/vissl/blob/master/vissl/losses/simclr_info_nce_loss.py#L127)). Do you want me to patch both at the same time?

indeed. I certainly welcome the improvements to SimCLR as well but in a separate PR :) thank you so much for this.

vissl/losses/barlow_twins_loss.py
prigoyal: makes sense. totally fine with keep it as is :)

vissl/losses/barlow_twins_loss.py
prigoyal: ah gotcha! I think we can go with `lambda_` for now since that way we will have a good correspondence with paper.

prigoyal:
@OlivierDehaene , I am pretty excited about this PR. I think we are extremely close to getting this merged and I appreciate you a lot for this work. :)

vissl/losses/barlow_twins_loss.py
```diff
@@ -0,0 +1,124 @@
...
+    def __repr__(self):
+        repr_dict = {
+            "name": self._get_name(),
+            "lambda_": self.loss_config.lambda_,
+            "embedding_dim": self.loss_config.embedding_dim
+        }
+        return pprint.pformat(repr_dict, indent=2)
+
+
+class BarlowTwinsCriterion(nn.Module):
```
QuentinDuval, line 59: First of all, thank you for the amazing work. All of this looks really really great 👍 

I just wanted to ask an open question more than a review comment (especially since I am not familiar with the initial implementation of @jzbontar) regarding the criterion:

It seems to me that the criterion defined here does not gather the representations across workers (in case of DistributedDataParallel training) which means that each GPU will compute its own cross-correlation matrix. Gathering the representation (as done in the SimCLR criterion) would increase the size of this cross-correlation matrix (by multiplying it with G, G being the number of GPUs used for the training).

So the open question (in particular for @jzbontar) is: should the criterion gather the representation (like SimCLR does for instance?) or is it not part of the Barlow Twins algorithm (and might actually harm the performance, or just increase the memory usage for no effect)?

OlivierDehaene: > It seems to me that the criterion defined here does not gather the representations across workers (in case of DistributedDataParallel training) which means that each GPU will compute its own cross correlation matrix, based on its own batch.

Indeed, this is not the case. I totally agree. It wasn't clear to me, so I opted for the naive implementation while waiting for the official implementation to come out.

> or just increase the memory usage for no effect

I am more afraid of the network overhead personally =/. Passing massive [Nx8192] matrices to the master process will have a non negligible effect on training speeds.

BTW, I was wondering if we could do something similar to Batch Norm process groups when computing the SimCLR loss to improve speeds. If we define one process group for each node of a distributed training, we could gather the embedding per process group, compute the loss for each process group and then aggregate the losses over the different groups. If the groups are large enough (meaning each node as multiple GPUs), this should not really impact the performances, shouldn't it? 

For example, if we have a distributed training of octo GPU nodes and each process has a batch size of 512, each process group would apply the SimCLR loss over a local batch size of 8*512 = 4096.

This should lead to speed gains as the large embedding are only shared between GPUs of the same node with very small overhead and we just share a small scalar between the nodes over the network.

OlivierDehaene: I am not sure if my rambling was very clear but I just thought about it when reading your comment and wanted your opinion on the subject. :)

QuentinDuval: > Indeed, this is not the case. I totally agree. It wasn't clear to me, so I opted for the naive implementation while waiting for the official implementation to come out.

It was not clear to me either :) Ultimately, I think we should rely on @jzbontar to provide us with the answer.

> I am not sure if my rambling was very clear but I just thought about it when reading your comment and wanted your opinion on the subject. :)

Your comment on SimCLR is quite interesting as well :)

In my understanding, the main goal of augmenting the number of nodes in SimCLR is to augment the total batch size that will go through the SimCLR criterion so that we can reach a big enough batch required to get good representations (in my past experiences, I usually selected the number of nodes such that I get the desired batch size). 

However, there is no real benefit in augmenting the batch size beyond 4096 or 8092 for SimCLR (and it seems to depend on the dataset), so past that point (if we increase the number of nodes further), we could indeed think of aggregating by process groups. This is indeed something that could be chosen by configuration.

jzbontar: That's a good point. In our code we synchronize values across all nodes and GPUs. In particular,

1. We use `torch.nn.SyncBatchNorm(..., affine=False)` when normalizing the columns of the embedding matrix
2. We use `torch.distributed.all_reduce(c)` to combine the cross-correlation matrix across all workers

I don't know how omitting these two steps affects the accuracy of Barlow Twins. Nevertheless, I think we should include both steps at first to be consistent with the paper. If it turns out that they are not important, we can remove them and have a per-gpu type loss instead.

OlivierDehaene: > We use `torch.distributed.all_reduce(c)` to combine the cross-correlation matrix across all workers

Oh I see! This is way better than gathering the embedding!

> I don't know how omitting these two steps affects the accuracy of Barlow Twins. Nevertheless, I think we should include both steps at first to be consistent with the paper.

Yes, I will patch the implementation when I come back from holidays :)

vissl/losses/barlow_twins_loss.py
```diff
@@ -0,0 +1,169 @@
...
+    """
+    The criterion corresponding to the Barlow Twins loss as defined in the paper
+    https://arxiv.org/abs/2103.03230v1.
+
+    Args:
+        lambda_ (float):        weight on the off-diagonal terms. It controls the trade-off
+                                between the importance given to the invariance term versus the
+                                redundancy reduction term.
+        scale_loss (float):     In order to match the code that was used to develop Barlow
```
OlivierDehaene, line 79: @jzbontar,

I saw in the official implementation that this parameter has been added. What purpose does it serve?

OlivierDehaene:
@prigoyal, @QuentinDuval,

The 2nd iteration of the criterion is done. The criterion now normalises the embedding and combines the cross-correlation matrix across all workers. 

I implemented the normalisation in `_sync_normalise` instead of using `torch.nn.SyncBatchNorm(..., affine=False)` to not have an unnecessary train/eval state in the criterion (since BatchNorm changes its behaviour).

@prigoyal, do I resolve the comments that are solved? Or do you do it? What is the process here between maintainers and contributors?

prigoyal:
> @prigoyal, @QuentinDuval,
> 
> The 2nd iteration of the criterion is done. The criterion now normalises the embedding and combines the cross-correlation matrix across all workers.
> 
> I implemented the normalisation in `_sync_normalise` instead of using `torch.nn.SyncBatchNorm(..., affine=False)` to not have an unnecessary train/eval state in the criterion (since BatchNorm changes its behaviour).
> 
> @prigoyal, do I resolve the comments that are solved? Or do you do it? What is the process here between maintainers and contributors?

thank you @OlivierDehaene , for the comments that pertain to the code changes here, contributors should resolve the comments once they have been addressed. If the comments require further discussion , they can stay open and we can discuss/maintainers can close :)

OlivierDehaene:
@jzbontar,

Forget me if I'm wrong but I thought that `torch.distributed.all_reduce` cut gradients. However you use it [here](https://github.com/facebookresearch/barlowtwins/blob/main/main.py#L208) to sum the cross-correlation matrices. Wouldn't that cause issues in the backward pass?

vissl/losses/barlow_twins_loss.py
jzbontar: The `--scale-loss` parameter multiplies the loss by a constant factor (1/32 by default). We include it in the code because we wanted to match our "research" code. I don't think scaling the loss is necessary, however, simply omitting the parameter and training with an unscaled loss results in a model that's a few percentage points worse than a model trained with a scaled loss. I'll try to remove the parameter from the code this week.

I think you can omit loss scaling from the VISSL implementation for now.

jzbontar:
> @jzbontar,
> 
> Excuse me if I'm wrong but I thought that `torch.distributed.all_reduce` cut gradients. However you use it [here](https://github.com/facebookresearch/barlowtwins/blob/main/main.py#L208) to sum the cross-correlation matrices. Wouldn't that cause issues in the backward pass?

Hm, good point. I checked whether `torch.distributed.all_reduce` cuts the gradients when I was writing code for Barlow Twins and I found that it doesn't and that the gradients were computed correctly. For example, when I run the following code on a machine with 2 GPUs

```python
import torch

def main_worker(gpu):
    torch.distributed.init_process_group(backend='nccl', init_method='tcp://localhost:12345', world_size=2, rank=gpu)
    torch.cuda.set_device(gpu)
    x = torch.ones(1).cuda().requires_grad_()
    xx = (gpu + 2) * x
    torch.distributed.all_reduce(xx)
    xx.backward()
    print(f'gpu={gpu}, grad={x.grad}')

if __name__ == '__main__':
    torch.multiprocessing.spawn(main_worker, nprocs=2)
```

it outputs

```
gpu=1, grad=tensor([3.], device='cuda:1')
gpu=0, grad=tensor([2.], device='cuda:0')
```

which seems okay to me.

The [PyTorch documentation](https://pytorch.org/docs/stable/distributed.html), however, suggests that the autograd-enabled communication primitives reside in `torch.distributed.nn`, so I probably should have used (and we should use in the VISSL implementation) `torch.distributed.nn.all_reduce` instead of `torch.distributed.all_reduce`.

Good catch, @odelalleau!

odelalleau:
> Good catch, @odelalleau!

Giving credit where it's due: good catch @OlivierDehaene! ;)

prigoyal:
@OlivierDehaene , thank you for your continued awesome work on this PR :) Just wanted to chime in for an early recommendation for a next step: in order for us to ensure reproducibility , we should aim for an entry with Barlow Twins in the VISSL Model zoo. This entry should have a pre-trained model on ImageNet (cc @jzbontar) and we should list the evaluation accuracy of this model in the table. 

For training on ImageNet, @jzbontar mentioned he might be able to help by running your code so it would be good to coordinate a plan for that. For the linear evaluation on ImageNet, please do reach out if you run into any issues or have any questions :)

configs/config/pretrain/barlow_twins/barlow_twins_4node_resnet.yaml
```diff
@@ -0,0 +1,121 @@
...
+      nesterov: False
+      num_epochs: 1000
+      regularize_bn: False
+      regularize_bias: False
+      param_schedulers:
+        lr:
+          auto_lr_scaling:
+            auto_scale: true
+            base_value: 0.2
+            base_lr_batch_size: 256
```
prigoyal, line 97: just wanted to double check:  do you mean to turn on the auto-scaling? if training on 32gpus (this config file), this LR value will be scaled to `(64 * 32 / 256) * 0.2 = 1.6` .

docs/source/ssl_approaches/barlow_twins.rst
```diff
@@ -0,0 +1,153 @@
...
+Stéphane Deny** in `this paper <https://arxiv.org/abs/2103.03230v1>`_.
+
+How to train Barlow Twins model
+-------------------------------
+
+VISSL provides yaml configuration file containing the exact hyperparam settings to
+reproduce the model. VISSL implements all the components including loss, data
+augmentations, collators etc required for this approach.
+
+To train ResNet-50 model on 2-machines (4-nodes) on ImageNet-1K dataset with Barlow Twins approach using MLP-head, lambda of 0.0051, scale loss of 0.024 and feature projection dimension 8192:
```
prigoyal, line 15: nit: 2 machines == 2 nodes ?

prigoyal, line 15: nit: for the MLP head, can we clarify whether 2-mlp, etc type?

tests/test_losses.py
```diff
@@ -69,6 +69,59 @@ def test_swav_loss(self):
         )
         _ = loss_layer(scores=self._get_embedding(), head_id=0)
 
+    def test_barlow_twins_loss(self):
```
prigoyal, line 72: nice! thank you :)

tests/test_losses.py
```diff
@@ -69,6 +69,59 @@ def test_swav_loss(self):
...
+            lambda_=0.0051, scale_loss=0.024, embedding_dim=EMBEDDING_DIM
+        )
+        embeddings = torch.randn((4, EMBEDDING_DIM), requires_grad=True)
+
+        self.assertTrue(embeddings.grad is None)
+        criterion(embeddings).backward()
+        self.assertTrue(embeddings.grad is not None)
+        with torch.no_grad():
+            next_embeddings = embeddings - embeddings.grad  # gradient descent
+            self.assertTrue(criterion(next_embeddings) < criterion(embeddings))
```
prigoyal, line 96: nice!

tests/test_losses.py
```diff
@@ -69,6 +69,59 @@ def test_swav_loss(self):
...
+            backend="nccl",
+            init_method="tcp://0.0.0.0:1234",
+            world_size=world_size,
+            rank=gpu_id,
+        )
+        criterion = BarlowTwinsCriterion(
+            lambda_=0.0051, scale_loss=0.024, embedding_dim=EMBEDDING_DIM
+        )
+        embeddings = torch.randn((batch_size, EMBEDDING_DIM),
+                                 dtype=torch.float32, requires_grad=True).cuda()
```
prigoyal, line 110: nit: formatting 
```
embeddings = torch.randn(
    (batch_size, EMBEDDING_DIM),
    dtype=torch.float32, 
    requires_grad=True
).cuda()
```

tests/test_losses.py
```diff
@@ -69,6 +69,59 @@ def test_swav_loss(self):
...
+        self.assertTrue(embeddings.grad is not None)
+        with torch.no_grad():
+            next_embeddings = embeddings - embeddings.grad  # gradient descent
+            self.assertTrue(criterion(next_embeddings) < criterion(embeddings))
+
+    @staticmethod
+    def worker_fn(gpu_id: int, world_size: int, batch_size: int):
+        dist.init_process_group(
+            backend="nccl",
+            init_method="tcp://0.0.0.0:1234",
```
prigoyal, line 102: I was wondering if we can use https://github.com/facebookresearch/vissl/blob/master/vissl/utils/misc.py#L104 instead to construct the run_id ? the issue  with hardcoding the port 1234 is that sometimes the machines could have the ports being used already and so the test could work on some machines but on the other machines it wouldn't .

vissl/losses/barlow_twins_loss.py
prigoyal: is the loss scaling determined and dependent on the #gpus ?

vissl/losses/barlow_twins_loss.py
```diff
@@ -0,0 +1,211 @@
...
+from classy_vision.losses import ClassyLoss, register_loss
+from torch import nn
+from torch.autograd.function import Function
+from vissl.utils.distributed_gradients import gather_from_all
+from vissl.utils.hydra_config import AttrDict
+
+
+class SyncNormalizeFunction(Function):
+    """
+    Adapted from: https://github.com/NVIDIA/apex/blob/master/apex/parallel/sync_batchnorm.py
```
prigoyal, line 16: curious question: why can't we use the function directly from NVIDIA? if we want to not introduce dependency on apex and that's the reason, that's valid and in that case we could introduce `vissl/utils/apex_helpers.py` and put this function there. This will allow usage beyond this loss function. :)

vissl/losses/barlow_twins_loss.py
```diff
@@ -0,0 +1,211 @@
...
+        self.lambda_ = lambda_
+        self.scale_loss = scale_loss
+        self.embedding_dim = embedding_dim
+        self.num_copies = 2
+        self.eps = 1e-5
+
+    @staticmethod
+    def _off_diagonal(x: torch.Tensor) -> torch.Tensor:
+        """
+        return a flattened view of the off-diagonal elements of a square matrix
```
prigoyal, line 152: nit: indentation is off

vissl/losses/barlow_twins_loss.py
```diff
@@ -0,0 +1,211 @@
...
+            dim=0,
+        )
+
+        # Do a gather over all embeddings, so we can compute the loss.
+        # Final shape is: (batch_size * num_gpus) x embedding_dim
+        embedding_normed_a = gather_from_all(embedding_normed_a)
+        embedding_normed_b = gather_from_all(embedding_normed_b)
+
+        # cross-correlation matrix
+        c = torch.mm(embedding_normed_a.T, embedding_normed_b) / (
```
prigoyal, line 193: nit: let's use `c` -> `correlation_matrix` something like that for more verbosity :)

prigoyal (commented):
looking really good. Small inline comments and thank you @OlivierDehaene for your hard work on this :)

vissl/losses/barlow_twins_loss.py
OlivierDehaene: For me there are several advantages to this:
- I think it is important that the criterion is not dependant on `train()` and `eval()` cycles. Using a BatchNorm layer would add hidden running statistics that could lead to weird behaviour. It is possible to disable said statistics using `track_running_stats =False`, however this argument is bugged in the Apex implementation of SyncBatchNorm. See my issue here: [https://github.com/NVIDIA/apex/issues/1071](https://github.com/NVIDIA/apex/issues/1071).
- We do not need to add additional logic to switch between PyTorch SyncBatchNorm and the faster Apex SyncBatchNorm depending on if Apex was installed by the user.

OlivierDehaene: > we could introduce vissl/utils/apex_helpers.py and put this function there. This will allow usage beyond this loss function. :)

100 % agree. I wasn't sure where to add it. It also needs to be tested.

OlivierDehaene:
@jzbontar,

I think there is still a weird interaction between `torch.all_reduce` and gradients. 

For example, the following code doesn't yield the same gradients when using `torch.all_reduce` and [`GatherLayer`](https://github.com/facebookresearch/vissl/blob/master/vissl/utils/distributed_gradients.py#L12), a specific gather op that doesn't cut gradients:

```python
import torch

from vissl.utils.distributed_gradients import GatherLayer

n, m = 8, 4

def main_worker(i):
    torch.distributed.init_process_group(backend='gloo', init_method='tcp://localhost:12345', world_size=4, rank=i)
    x = torch.full((n, m), i+1, dtype=torch.float32).requires_grad_()

    # compute xTx using all_reduce
    xTx = x.T @ x
    torch.distributed.all_reduce(xTx)
    xTx.sum().backward()
    if i == 0:
        print(x.grad)

# tensor([[8., 8., 8., 8.],
#         [8., 8., 8., 8.],
#         [8., 8., 8., 8.],
#         [8., 8., 8., 8.],
#         [8., 8., 8., 8.],
#         [8., 8., 8., 8.],
#         [8., 8., 8., 8.],
#         [8., 8., 8., 8.]])

    x.grad.zero_()

    # compute xTx using GatherLayer
    z = torch.cat(GatherLayer.apply(x))
    zTz = z.T @ z
    zTz.sum().backward()
    if i == 0:
        print(x.grad)

# tensor([[32., 32., 32., 32.],
#         [32., 32., 32., 32.],
#         [32., 32., 32., 32.],
#         [32., 32., 32., 32.],
#         [32., 32., 32., 32.],
#         [32., 32., 32., 32.],
#         [32., 32., 32., 32.],
#         [32., 32., 32., 32.]])

if __name__ == '__main__':
    torch.multiprocessing.spawn(main_worker, nprocs=4)

```

Could this explain your scaling issue?

configs/config/pretrain/barlow_twins/barlow_twins_4node_resnet.yaml
OlivierDehaene: Yes, auto scaling should be turned on. See: https://github.com/facebookresearch/barlowtwins/blob/52ffe5672cce03761036b401651cdf8459dcb72e/main.py#L149

jzbontar:
> EDIT: it may actually be an error in GatherLayer. @prigoyal, @QuentinDuval, shouldn't there be a rescaling by 1/world_size here?

Yes, right? I think so too. The following example shows that the `torch.distributed.all_reduce` solution computes gradients that are the same as the baseline one-process solution. Whereas the loss in the `torch.distributed.nn.all_reduce` solution needs to be scaled by `world_size`.

```python
import torch
import torch.distributed.nn

# compute gradient of xTy**2
# the gradient wrt x is 2 * xTy * y
# the gradient wrt y is 2 * xTy * x

def main_worker(rank):
    torch.distributed.init_process_group(backend='gloo', init_method='tcp://localhost:12345', world_size=4, rank=rank)

    # torch.distributed.all_reduce solution
    x = torch.randn(4).requires_grad_()
    y = torch.randn(4).requires_grad_()
    xTy = torch.dot(x, y)
    torch.distributed.all_reduce(xTy)
    loss = xTy.pow(2)
    loss.backward()
    assert torch.equal(x.grad, 2 * xTy * y) # this is correct and matches the baseline
    assert torch.equal(y.grad, 2 * xTy * x)

    # torch.distributed.nn.all_reduce solution
    x = torch.randn(4).requires_grad_()
    y = torch.randn(4).requires_grad_()
    xTy_1gpu = torch.dot(x, y)
    xTy = torch.distributed.nn.all_reduce(xTy_1gpu)
    loss = xTy.pow(2)
    loss /= 4  # have to divide loss by world_size to match the baseline solution
    loss.backward()
    assert torch.equal(x.grad, 2 * xTy * y)
    assert torch.equal(y.grad, 2 * xTy * x)

if __name__ == '__main__':
    # single process baseline
    x = torch.randn(16).requires_grad_()
    y = torch.randn(16).requires_grad_()
    xTy = torch.dot(x, y)
    loss = xTy.pow(2)
    loss.backward()
    assert torch.equal(x.grad, 2 * xTy * y)
    assert torch.equal(y.grad, 2 * xTy * x)

    torch.multiprocessing.spawn(main_worker, nprocs=4)
```

Maybe we should really be using `reduce` instead of `all_reduce`?

prigoyal:
> @prigoyal, @QuentinDuval, shouldn't there be a rescaling by `1/world_size` here? Since the default reduce op is SUM. https://github.com/facebookresearch/vissl/blob/master/vissl/utils/distributed_gradients.py#L27

@OlivierDehaene , in VISSL, https://github.com/facebookresearch/vissl/blob/master/vissl/trainer/train_steps/standard_train_step.py#L158 is the place where we call `all_reduce_mean` from classy vision which divides by the world size https://github.com/facebookresearch/ClassyVision/blob/master/classy_vision/generic/distributed_util.py#L69-L73

QuentinDuval:
> EDIT: it may actually be an error in GatherLayer.
> @prigoyal, @QuentinDuval, shouldn't there be a rescaling by `1/world_size` here? Since the default reduce op is SUM. https://github.com/facebookresearch/vissl/blob/master/vissl/utils/distributed_gradients.py#L27

I do not see any issue with `GatherLayer` in itself: the implementation sums the gradients coming from all workers, but then fetch only the parts applicable to the given worker (to the inputs coming from this worker). So each "sum" is only used once, there is no double counting of any gradient contribution.

Originally, this implementation is there to deal with things like SimCLR loss, in which each worker compute a different part of the loss. It is true thought, that compared to a `all_gather` without gradient propagation, then the contribution will be `world_size` times bigger. Is it why you think we need to downscale it?

OlivierDehaene:
> It is true thought, that compared to a all_gather without gradient propagation, then the contribution will be world_size times bigger. Is it why you think we need to downscale it?

I might be confused but I don't understand why running on multiple processes should influence the value of the gradient.

If I distribute a batch over two processes, I do not expect each process to receive twice the amount of gradient. Or, I think that is what is happening here.

```python
    @staticmethod
    def backward(ctx, *grads):
        #grads: List[torch.Tensor] with grads[i].shape == Nx... and len(grads) == world_size
        all_gradients = torch.stack(grads) # WORLDSIZExNx...
        dist.all_reduce(all_gradients) # sum all_gradients over all processes
        return all_gradients[dist.get_rank()] # return correct value BUT multiplied by number of processes due to sum above
```

I am missing something here? Is it actually a good idea to multiply the gradient by the number of processes?

QuentinDuval:
> > It is true thought, that compared to a all_gather without gradient propagation, then the contribution will be world_size times bigger. Is it why you think we need to downscale it?
> 
> I might be confused but I don't understand why running on multiple processes should influence the value of the gradient.
> 
> If I distribute a batch over two processes, I do not expect each process to receive twice the amount of gradient. Or, I think that is what is happening here.
> 
> I am missing something here? Is it actually a good default to multiply the gradient by the number of processes?

Let me do some checks on what's exactly happening and I will come back to you. I feel like I am the one missing something :)

OlivierDehaene:
> The reason is not that the GatherLayer is wrong in terms of what it does (it does not multiply the gradients, it simply takes into account the contributions of all workers), but that it is not aligned with the way PyTorch computes gradients (PyTorch does not sum the gradients, but averages them).

Yes, in my case it was multiplied since all workers had the same gradients.

However, we might be missing a piece of the puzzle as the pytorch team does exactly the same thing in their implementation of `all_gather`: https://github.com/pytorch/pytorch/blob/2f5db687977aea95d1b0abb8a78f9a494640114a/torch/distributed/nn/functional.py#L223.

QuentinDuval:
> > The reason is not that the GatherLayer is wrong in terms of what it does (it does not multiply the gradients, it simply takes into account the contributions of all workers), but that it is not aligned with the way PyTorch computes gradients (PyTorch does not sum the gradients, but averages them).
> 
> Yes, in my case it was multiplied since all workers had the same gradients.
> 
> However, we might be missing a piece of the puzzle as the pytorch team does exactly the same thing in their implementation of `all_gather`: https://github.com/pytorch/pytorch/blob/2f5db687977aea95d1b0abb8a78f9a494640114a/torch/distributed/nn/functional.py#L223.

Yes, I think the implementation is correct (see **[EDIT 2]** in my answer above): it is just that we need to average the losses and not sum them (which is what happens now, because each worker computes its loss and adds its contribution to the gradients VS the loss would be averaged if the work was done on a single worker).

OlivierDehaene:
> Yes, I think the implementation is correct (see **[EDIT 2]** in my answer above): it is just that we need to average the losses and not sum them (which is what happens now, because each worker computes its loss and adds its contribution to the gradients VS the loss would be averaged if the work was done on a single worker).

As @prigoyal pointed out, the loss is already average over the workers: 

> @OlivierDehaene , in VISSL, https://github.com/facebookresearch/vissl/blob/master/vissl/trainer/train_steps/standard_train_step.py#L158 is the place where we call all_reduce_mean from classy vision which divides by the world size https://github.com/facebookresearch/ClassyVision/blob/master/classy_vision/generic/distributed_util.py#L69-L73

OlivierDehaene:
Ok I see now! Since the last `torch.distributed.all_reduce` operation cuts gradients, everything fall into place! 

```python
import torch
import torch.distributed.nn

from vissl.utils.distributed_gradients import GatherLayer

# compute gradient of xTy**2
# the gradient wrt x is 2 * xTy * y
# the gradient wrt y is 2 * xTy * x

def main_worker(rank):
    torch.distributed.init_process_group(backend='gloo', init_method='tcp://localhost:12345', world_size=4, rank=rank)

    # torch.distributed.nn.all_gather solution
    x = torch.randn(8).requires_grad_()
    y = torch.randn(8).requires_grad_()
    gather_x = torch.cat(GatherLayer.apply(x))
    gather_y = torch.cat(GatherLayer.apply(y))
    xTy = torch.dot(gather_x, gather_y)
    loss = xTy.pow(2)

    loss = all_reduce_mean(loss)

    loss.backward()
    assert torch.equal(x.grad, 2 * xTy * y)
    assert torch.equal(y.grad, 2 * xTy * x)

if __name__ == '__main__':
    # single process baseline
    x = torch.randn(16).requires_grad_()
    y = torch.randn(16).requires_grad_()
    xTy = torch.dot(x, y)
    loss = xTy.pow(2)
    loss.backward()
    assert torch.equal(x.grad, 2 * xTy * y)
    assert torch.equal(y.grad, 2 * xTy * x)

    torch.multiprocessing.spawn(main_worker, nprocs=4)
```

OlivierDehaene:
@prigoyal, @QuentinDuval,

Then I have another remark. In the case of SimCLR and Barlow Twins (not sure about the others), the loss is the same between all workers meaning that we don't need to [reduce over all workers](https://github.com/facebookresearch/vissl/blob/master/vissl/trainer/train_steps/standard_train_step.py#L158). Removing this op in those cases saves one round of all-to-all communication per step. 

Do you feel that it would be worth it to find a way to remove this reduce in those cases? Or it would add too much complexity for a marginal gain. 

Cheers for all the interesting conversations btw, this is really fun :)
Tell me if you have another channel to have this kind of talks.

OlivierDehaene:
@jzbontar, @prigoyal, @QuentinDuval,

I launched a full training on ImageNet on our cluster. See you in 4 days =D

QuentinDuval:
> @jzbontar, @prigoyal, @QuentinDuval,
> 
> I launched a full training on ImageNet on our cluster. See you in 3 days =D

Awesome !!! 👍

OlivierDehaene:
`Infinite Loss or NaN` at epoch 62.

Damnit...

Also, the loss value seems abnormally high. 

```
INFO 2021-03-26 06:37:21,842 log_hooks.py: 208: Rank: 0; [ep: 62] iter: 38760; lr: 1.58488; loss: 72.75297; btime(ms): 622; eta: 4 days, 5:19:40; peak_mem: 10339M; btime(313iters): 677 ms; rolling_eta: 4 days, 14:16:43
```

Where as from the [original logs](https://dl.fbaipublicfiles.com/barlowtwins/epochs1000_bs2048_lr0.2_lambd0.0051_proj_8192_8192_8192_scale0.024/stats.txt) in @jzbontar runs we see at epoch 62:
```
{"epoch": 62, "step": 38900, "learning_rate": 1.5890852084050442, "repr_loss": 12.968552589416504, "time": 36789}
```

No free lunch I guess, back to debugging.

prigoyal:
@OlivierDehaene , I'm curious what the loss curves of both Jure's and your runs look like if we compare them. Additionally plot the learning rate value to compare. This might give us some idea too :) just a thought.

OlivierDehaene:
image

image

What is weird to me is that I benchmarked the VISSL loss implementation vs the official one on Imagenette 160 and didn't see a discrepancy. Maybe the bug appears with > 1 nodes as I only did so on my home setup with a single card.

prigoyal:
thank you for sharing the curves @OlivierDehaene , it indeed seems to me that the issue could be in multi-node case. Thinking out loud: there are 2 issues looks like - 
1) loss value is high 4x -> is this the case also when the you train on 1node ? also the `scale_loss` comes to mind now. I am not sure what the importance of this parameter it but do we adjust it as we change the #nodes ?

2) the loss become NaN -> given how the loss has decreased, it seems to me that it could be due to regularization. I have observed similar things but in context of SwAV trainings and the solution there was to use a better weight decay value. To debug this, I'd recommend to plot the stats (mean, max, std etc) for the activations of the model as the training progresses. This should allow to see that regularization is indeed the issue.

OlivierDehaene:
I did a small test using the official loss implementation but in VISSL and faced the same issue so it leads me to believe that it might be either a data augmentation or an optimizer problem.

prigoyal:
Hi @OlivierDehaene , hope things are well :) just wanted to check in if you needed any help from us in any way?

OlivierDehaene:
> Hi @OlivierDehaene , hope things are well :) just wanted to check in if you needed any help from us in any way?

Hello @prigoyal, I didn't find time to work on this at all this week. I will continue to investigate as soon as possible.

jingli9111:
Hi @OlivierDehaene,

I would like to help with the NaN problem. 
When we were developing Barlow twins, we also had NaN problem when training with SGD.
Our standard training recipe is to use LARS optimizer, exclude parameters from BNs.

I see in your code, the optimizer is set as `sgd` and `use_larc: true`. So this should be right.
But do you have specific lines to exclude BN parameters? I guess this might be the problem.

OlivierDehaene:
> But do you have specific lines to exclude BN parameters? I guess this might be the problem.

Hello @jingli9111! No, BN parameters are not excluded at the moment. We could add them to the list of `non_regularized_parameters` in the training config. 

Thanks for the tip! I will try it asap.

OlivierDehaene:
> But do you have specific lines to exclude BN parameters? I guess this might be the problem.

Hello @jingli9111! No, BN parameters are not excluded at the moment. We could add them to the list of `NON_TRAINABLE_PARAMS` in the training config.

Thanks for the tip! I will try it asap.

prigoyal:
> > But do you have specific lines to exclude BN parameters? I guess this might be the problem.
> 
> Hello @jingli9111! No, BN parameters are not excluded at the moment. We could add them to the list of `NON_TRAINABLE_PARAMS` in the training config.
> 
> Thanks for the tip! I will try it asap.
> 
> EDIT: I misunderstood your message. You only exclude them from LARS but still train them. I don't know if this is possible in VISSL. @prigoyal, any ideas?
> 
> However, I don't see how this would have such a drastic effect on the training loss. So there might be an additional issue.

@jingli9111 , when you say exclude the BN parameters, can you clarify what exactly does that mean? we use the LARC from apex https://github.com/NVIDIA/apex/blob/master/apex/parallel/LARC.py which doesn't offer other explicit ways to exclude the parameters.

jingli9111:
@OlivierDehaene  You are right. The BN parameters are only excluded from LARS but they still need to be trained. They just don't go through the LARS gradient adaptation. You can see detail here: https://github.com/facebookresearch/barlowtwins/blob/633dc560684039f5b52822fb4085144f75b7c65d/main.py#L235

@prigoyal  See above, it means that the BN parameters are not adapted by the optimizer. They are only updated by original gradient. 
Sorry I don't know how LARC from apex can exclude parameters. Possible solution:
LARC has a function of `add_param_group`, maybe we can manually combine LARC with part of the parameters and a standard SGD on the rest... No sure if this is possible in VISSL's framework.

prigoyal:
> @OlivierDehaene You are right. The BN parameters are only excluded from LARS but they still need to be trained. They just don't go through the LARS gradient adaptation. You can see detail here: https://github.com/facebookresearch/barlowtwins/blob/633dc560684039f5b52822fb4085144f75b7c65d/main.py#L235
> 
> @prigoyal See above, it means that the BN parameters are not adapted by the optimizer. They are only updated by original gradient.
> Sorry I don't know how LARC from apex can exclude parameters. Possible solution:
> LARC has a function of `add_param_group`, maybe we can manually combine LARC with part of the parameters and a standard SGD on the rest... No sure if this is possible in VISSL's framework.

Looking at the LARS in BarlowTwins code, a few things I can spot:
1. it seems like the bias and BN parameters are not regularized (https://github.com/facebookresearch/barlowtwins/blob/633dc560684039f5b52822fb4085144f75b7c65d/main.py#L92 -> here the weight decay is set to 0). In VISSL, this doable by setting the https://github.com/facebookresearch/vissl/blob/master/vissl/config/defaults.yaml#L819-L822 -> @OlivierDehaene can you verify if you are already training with these settings?
2. likewise for bias and BN parameters, it seems like the LARS is excluded . @jingli9111 , we can construct the optimizer params groups however, fundamentally, it seems like one needs to change the LARC implementation instead to not apply the LARC on the bias + BN parameters. I don't see how one can simply control this otherwise from simply `param_groups` (but maybe I am missing something?). 

@OlivierDehaene , if we really need the 2) option above, my recommendation is that we add a `LARS` optimizer to VISSL that looks like https://github.com/NVIDIA/apex/blob/master/apex/parallel/LARC.py except that we modify the `step` function + optionally take the parameters `regularize_bn, regularize_bias`  and use it. LMK what you think? It would definitely be good to verify that 2) is a must option and that 1) itself is not enough.

OlivierDehaene:
>  @OlivierDehaene can you verify if you are already training with these settings?

@prigoyal Yes, they are already disabled. See: https://github.com/facebookresearch/vissl/pull/230/files#diff-410574be774d8e11fe23936885406492e1cb1f225f2d8909a8e0145126f34114R90

prigoyal:
> > @OlivierDehaene can you verify if you are already training with these settings?
> 
> @prigoyal Yes, they are already disabled. See: https://github.com/facebookresearch/vissl/pull/230/files#diff-410574be774d8e11fe23936885406492e1cb1f225f2d8909a8e0145126f34114R90

thank you for confirming @OlivierDehaene . I connected with @jingli9111 and he's running a few tests to verify that the LARS optimizer is indeed required. If that's the case, we will have to take the route 2) and do the 2 steps: a) we implement `vissl/optimizers/lars.py` https://github.com/facebookresearch/vissl/tree/master/vissl/optimizers and b) also, since we use the SGD from ClassyVision , we will have to create `sgd_lars.py` that inherits https://github.com/facebookresearch/ClassyVision/blob/master/classy_vision/optim/sgd.py#L42 and overrides the `prepare` function. lmk if this sounds

jingli9111:
Screen Shot 2021-04-07 at 8 39 10 AM
Here's what I find. The blue curve is the experiment of "not exclude BN parameters for LARS adaptation". You can see the loss is going up. It may or may not in the end drive to NaN. But it's clear that this "parameters exclusion" setting is crucial to get expected performance.
I am wondering if this is the same as your NaN problem @OlivierDehaene. Your NaN problem is a sudden jump, right? Then it might be caused by something else.

OlivierDehaene:
> But it's clear that this "parameters exclusion" setting is crucial to get expected performance.

@jingli9111,
Thanks for your experiments! I added [your LARS implementation](https://github.com/facebookresearch/vissl/pull/230/files#diff-67a5e00aa664c919efd379d888e42e3542617dec3c082cff1659bed0ccdbac69R58) to the PR if you want to review it.

Since in VISSL, there is already different groups for the parameters that need to be regularized and the others, we only need to filter for the LARS adaptive LR. See https://github.com/facebookresearch/vissl/blob/master/vissl/optimizers/optimizer_helper.py#L102

jingli9111:
A quick update: The blue curve above (not exclude BN parameters for LARS adaptation) gets NaN.

prigoyal:
> A quick update: The blue curve above (not exclude BN parameters for LARS adaptation) gets NaN.

this is great news! @OlivierDehaene , have you tried running with LARS and does that help with your training?

prigoyal:
@OlivierDehaene , please checkout https://github.com/facebookresearch/vissl/pull/288 for a one line change related to bias parameter regularization.

jingli9111:
[log.txt](https://github.com/facebookresearch/vissl/files/6297344/log.txt)

I am able to run @OlivierDehaene 's branch. See log file above. Looks like the loss matches the figure above (loss ~70) and there's no NaN yet (158 epochs, 100k steps)

There's a small discrepancy here different from the master branch that may cause problem.  https://github.com/OlivierDehaene/vissl/blob/29171574c23df59f5f617964ee8e5a5c718c9ab0/vissl/trainer/train_task.py#L460

prigoyal:
thank you @jingli9111 .

@OlivierDehaene , please let us know on what you think could be the remaining things you were planning. Li is running the training on ImageNet from your branch and we can test his trained model on downstream task. Once the evaluation comes out fine, we are ready to merge the PR. 

One action item for you that I can identify is: Rebase this PR on master branch of VISSL and update. 

Let me know what you think. We would love to merge the Barlow Twins implementation in VISSL master by end of April 16 (does that date work)?

OlivierDehaene:
@prigoyal,

That's great! There is still a bit of documentation missing but otherwise if the results come out ok, it is ready. 
However, does the current training reflect the #288 changes?

prigoyal:
> @prigoyal,
> 
> That's great! There is still a bit of documentation missing but otherwise if the results come out ok, it is ready.
> However, does the current training reflect the #288 changes?

@jingli9111 , can you confirm the #288 changes were used or not in the code? :)

jingli9111:
> > @prigoyal,
> > That's great! There is still a bit of documentation missing but otherwise if the results come out ok, it is ready.
> > However, does the current training reflect the #288 changes?
> 
> @jingli9111 , can you confirm the #288 changes were used or not in the code? :)

No, I didn't include #288. 
So it may give different results.
I will wait until current experiment finish, just to verify there's no bug. And then I can rerun to include new changes.

jingli9111:
The original experiment does not have a bug. But the accuracy seems lower than expected.
I am starting running Olivier's new code. It's working well now.

prigoyal:
thank you for the update @jingli9111  :) and glad to know the new code is working well. keep us posted :)

prigoyal:
Hi @OlivierDehaene , @jingli9111 , just wanted to check-in how things are going in regards to this PR and if there are blockers that I can help with? :)

jingli9111:
The current code works well and there's no NaN problem anymore.
However, it gives worse results (65.2% for 1000 epoch running). 
Looks like it is caused by the loss scale.

https://github.com/OlivierDehaene/vissl/blob/e3a74b67b47c536d6edec52db26db446f3fb1bc2/vissl/losses/barlow_twins_loss.py#L135

We originally used `reduced_sum` with entire batch size. Here you used `all_reduce_mean` with `per gpu batch size`.
Mathematically the forward pass should be the same. But will it make the gradient different?

Also, the per gpu batch size seems need to be divided by 2. Because you split the tensor after calculate the batch size.  


I hard coded the change like this
```
        # cross-correlation matrix
        correlation_matrix = torch.mm(embedding_normed_a.T, embedding_normed_b) / (batch_size * 32 / 2)

        # Reduce cross-correlation matrices from all processes
        correlation_matrix = all_reduce_sum(correlation_matrix)
```
and the now loss seems a perfect match.
loss

OlivierDehaene:
@jingli9111,

Thank you for all your experiments! Does this mean that we can get rid of the `scale_loss` parameter?

OlivierDehaene:
@jingli9111,

Either way I'm happy to change to your version and get this over with. =D

prigoyal:
> @jingli9111,
> 
> Either way I'm happy to change to your version and get this over with. =D

thank you @OlivierDehaene , I do agree with your sentiment that the impact on training is not clear especially since mathematically it all seems equal. I think we should debug this later to understand "how does it impact the gradients". 

For now, I suggest that we move forward with trying to land this PR . 

@OlivierDehaene , would you be able to update the PR accordingly?

@jingli9111 , is it possible to add the model you are training to the model zoo in VISSL? (feel free to ping me and we can coordinate). It seems like for the linear probe, we need to use the same exact transfer setup you use. Let's coordinate about this too.

jingli9111:
@OlivierDehaene 

Sorry I didn't make it clear.
Yeah, you are right. Using `all_reduce_sum/world_size` should be the same.

I was using `world_size=32`. 
So the only difference is that `/2`. 

The batch size is calculated before you split the embedding tensor: `https://github.com/OlivierDehaene/vissl/blob/e3a74b67b47c536d6edec52db26db446f3fb1bc2/vissl/losses/barlow_twins_loss.py#L116` 

So I think it should be simply changed to 
```
correlation_matrix = torch.mm(embedding_normed_a.T, embedding_normed_b) / (batch_size / 2)

correlation_matrix = all_reduce_mean(correlation_matrix)
```

jingli9111:
loss
The current training loss curve (orange curve) update matches perfectly to the expected training loss (blue curve). I am also running linear probe on a shorter training model (300 epoch) and looks like the accuracy is good.

So I think it's ready to land the PR.

@prigoyal 
In terms of the model zoo, we can upload the 300 epoch model and update it with the 1000 epoch model later. Also, do we need to wait until the accuracy is verified?

prigoyal:
sounds amazing! I see the following follow-ups for us all here to facilitate coordination:

**Step1:** @jingli9111 

- [ ] can you push the linear eval yaml config to this PR or share with Olivier (via a gist.github.com for example)?
- [ ] can you share the model with me (ping me on chat) and we can upload the model to VISSL model zoo.
- [ ] can you report the accuracy of the model after linear probe? we will need to mention the accuracy in the model_zoo.md in VISSL

**Step2:** @OlivierDehaene ,

- [ ] will you be able to incorporate the comment from @jingli9111 to reflect the change he suggested in loss?
- [ ] Once @jingli9111 shares the linear eval and model zoo accuracy and model zoo url that I will share?
- [ ] afterwards, can you incorporate linear eval config, update the model zoo accuracy, url and rebase the PR?

**Step3:** @prigoyal  Review and land

vissl/utils/apex_helpers.py
```diff
@@ -0,0 +1,59 @@
+import torch
```
prigoyal, line 1: nit: would you be comfortable adding the license header here
```
# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
```

vissl/optimizers/lars.py
```diff
@@ -0,0 +1,112 @@
+from typing import Any, Dict
```
prigoyal, line 1: nit: would you be comfortable adding the license header here
```
# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
```

vissl/losses/barlow_twins_loss.py
```diff
@@ -0,0 +1,152 @@
...
+
+        # split embedding between copies
+        embedding_normed_a, embedding_normed_b = torch.split(
+            embedding_normed,
+            split_size_or_sections=batch_size // self.num_copies,
+            dim=0,
+        )
+
+        # cross-correlation matrix
+        correlation_matrix = torch.mm(embedding_normed_a.T, embedding_normed_b) / batch_size
```
prigoyal, line 132: noting the change recommended by @jingli9111 
```
correlation_matrix = torch.mm(embedding_normed_a.T, embedding_normed_b) / (batch_size / 2)
```

MODEL_ZOO.md
```diff
@@ -205,3 +206,9 @@ There is some standard deviation in linear results if we run the same eval sever
 | Method | Model | PreTrain dataset | ImageNet top-1 acc. | URL |
 | ------ | ----- | ---------------- | ------------------- | --- |
 | [MoCo-v2](https://arxiv.org/abs/2003.04297)   |    RN50 - 200 epochs - 256 batch-size         |    ImageNet-1K      | 66.4 | [model](https://dl.fbaipublicfiles.com/vissl/model_zoo/moco_v2_1node_lr.03_step_b32_zero_init/model_final_checkpoint_phase199.torch)
+
+### BarlowTwins
+
+| Method | Model | PreTrain dataset | ImageNet top-1 acc. | URL |
+| ------ | ----- | ---------------- | ------------------- | --- |
+| [Barlow Twins](https://arxiv.org/abs/2103.03230)   |    RN50 - 1000 epochs - 2048 batch-size         |    ImageNet-1K      | #TODO | #TODO
```
prigoyal, line 214: @jingli9111 has trained a model using your PR on Imagenet

Model specification change: `RN50 - 300 epochs - 2048 batch-size`

Model path: `https://dl.fbaipublicfiles.com/vissl/model_zoo/barlow_twins/barlow_twins_32gpus_4node_imagenet1k_300ep_resnet50.torch`

Model accuracy: `70.75`

prigoyal (commented):
this looks ready to go. Just some minor thoughts inline + rebase on VISSL master :) thank you for all your efforts! :)

OlivierDehaene:
@prigoyal,

It should be good!

prigoyal:
ah , it requires a rebase, it's hitting an issue that was fixed in master https://github.com/facebookresearch/vissl/blob/master/vissl/trainer/train_task.py#L462 :)

OlivierDehaene:
@prigoyal,

I forgot I was on a fork and rebased on my fork instead of this one... Should be good now!

OlivierDehaene:
@prigoyal,

I can't see the details of the failing test so IDK what to patch.

prigoyal:
> @prigoyal,
> 
> I can't see the details of the failing tests so IDK what to patch.

Hi @OlivierDehaene ,  I will comment on relevant places in the PR shortly :)

docs/source/ssl_approaches/barlow_twins.rst
```diff
@@ -0,0 +1,134 @@
...
+    python tools/run_distributed_engines.py config=pretrain/barlow_twins/barlow_twins_4node_resnet \
+        config.MODEL.TRUNK.NAME=resnet config.MODEL.TRUNK.TRUNK_PARAMS.RESNETS.DEPTH=101
+
+* **Train ResNet-50-w2 (2x wider):**
+
+.. code-block:: bash
+
+    python tools/run_distributed_engines.py config=pretrain/barlow_twins/barlow_twins_4node_resnet \
+        config.MODEL.TRUNK.NAME=resnet config.MODEL.TRUNK.TRUNK_PARAMS.RESNETS.DEPTH=101 \
+        config.MODEL.TRUNK.TRUNK_PARAMS.RESNETS.WIDTH_MULTIPLIER=2
```
prigoyal, line 86: we recently removed `TRUNK_PARAMS` from VISSL

configs/config/debugging/pretrain/barlow_twins/barlow_twins_1node_resnet_imagenette_160.yaml
```diff
@@ -0,0 +1,119 @@
...
+    TRUNK:
+      NAME: resnet
+      TRUNK_PARAMS:
+        RESNETS:
+          DEPTH: 50
+    HEAD:
+      PARAMS: [
+        ["mlp", {"dims": [2048, 8192], "use_relu": True, "use_bn": True, "use_bias": False}],
+        ["mlp", {"dims": [8192, 8192], "use_relu": True, "use_bn": True, "use_bias": False}],
+        ["mlp", {"dims": [8192, 8192], "use_bias": False}],
```
prigoyal, line 70: We've recently made an improvement in VISSL where one can construct the mlps like this
```
["mlp", {"dims": [2048, 8192, 8192, 8192], "use_relu": True, "use_bn": True, "use_bias": False}],
```

However, since the model trained by @jingli9111  in the model zoo follows the old head settings, the change required is:
```
["mlp", {"dims": [2048, 8192], "use_relu": True, "use_bn": True, "use_bias": False, "skip_last_layer_relu_bn": False}],
["mlp", {"dims": [8192, 8192], "use_relu": True, "use_bn": True, "use_bias": False, "skip_last_layer_relu_bn": False}],
["mlp", {"dims": [8192, 8192], "use_bias": False}],
```

configs/config/pretrain/barlow_twins/barlow_twins_4node_resnet.yaml
```diff
@@ -0,0 +1,116 @@
...
+  MODEL:
+    TRUNK:
+      NAME: resnet
+      TRUNK_PARAMS:
+        RESNETS:
+          DEPTH: 50
+    HEAD:
+      PARAMS: [
+        ["mlp", {"dims": [2048, 8192], "use_relu": True, "use_bn": True}],
+        ["mlp", {"dims": [8192, 8192], "use_relu": True, "use_bn": True}],
```
prigoyal, line 63:
```
["mlp", {"dims": [2048, 8192], "use_relu": True, "use_bn": True, "skip_last_layer_relu_bn": False}],
["mlp", {"dims": [8192, 8192], "use_relu": True, "use_bn": True, "skip_last_layer_relu_bn": False}],
```

configs/config/test/integration_test/quick_barlow_twins.yaml
```diff
@@ -0,0 +1,116 @@
...
+  MODEL:
+    TRUNK:
+      NAME: resnet
+      TRUNK_PARAMS:
+        RESNETS:
+          DEPTH: 50
+    HEAD:
+      # Reduced MLP size to limit memory footprint
+      PARAMS: [
+        ["mlp", {"dims": [2048, 128], "use_relu": True, "use_bn": True, "use_bias": False}],
```
prigoyal, line 65:
```
["mlp", {"dims": [2048, 128], "use_relu": True, "use_bn": True, "use_bias": False, "skip_last_layer_relu_bn": False}],
```

prigoyal (commented):
I left some last remaining inline comments

Additionally, we need to run linter on the PR. Follow the instructions https://github.com/facebookresearch/vissl/blob/master/dev/README.md#practices-for-coding-quality to run the linter provided by VISSL and update the PR :)

docs/source/ssl_approaches/barlow_twins.rst
OlivierDehaene: It seems that all docs in `ssl_approaches` are outdated.

OlivierDehaene: I don't really know how to update it. The docs are mainly a copy/paste of the `sim_clr` one

prigoyal: hmmm, we did update our docs as well to reflect it. On quick search in VISSL master, https://github.com/facebookresearch/vissl/search?q=TRUNK_PARAMS I am not seeing the TRUNK_PARAMS in docs. 

What master are you at ? :)

prigoyal: I can take care of updating the docs later in a follow-up PR :) we can proceed with this PR without doc updates! :)

OlivierDehaene: It should be good but I didn't patch the other docs.

OlivierDehaene:
@prigoyal,

Some lints fail on file that were not touched by this PR. I left them as is.

prigoyal:
> @prigoyal,
> 
> Some lints fail on file that were not touched by this PR. I left them as is.

@OlivierDehaene , that sounds great. I'll work on linting the other files :)

docs/source/ssl_approaches/barlow_twins.rst
OlivierDehaene: @prigoyal, 

See this file: https://github.com/facebookresearch/vissl/blob/master/docs/source/ssl_approaches/simclr.rst#training-different-model-architecture

OlivierDehaene: Github search is not reliable in my experience.

OlivierDehaene: Should I only remove `TRUNK_PARAMS`? I thought I was suppose to remove both `TRUNK` and `TRUNK_PARAMS`.

OlivierDehaene: OK I misunderstood you. I will add `TRUNK` back =P

OlivierDehaene:
Sorry I misunderstood what you meant with `TRUNK` and `TRUNK_PARAMS`. It should be good now!

PR #231, opened 2021-03-11
[fix] Full tuning configurations for imagenet and places205

QuentinDuval:
The evaluation mode was not enabled, leading to a crash when loading the weights of a pre-trained model. The fix only consists  in enabling the evaluation.


prigoyal:
thank you @QuentinDuval :) can we double check that the other config files for full finetuning are unaffected as well. :) otherwise let's import this to phabricator and merge it :)

prigoyal:
Hi @QuentinDuval , let's aim to land this PR to ensure the benchmarks work smoothly. Can you import the PR to Phabricator ? :)

PR #232, opened 2021-03-12
[Benchmark] Evaluating robustness of learned representations

QuentinDuval:
This is a first draft at a benchmark for the robustness to distribution shifts or the learned representation, so that I can get your feedback on it (Is it a good idea? Is there a way to make this fit better with the rest of VISSL?) as well as triggering some discussions.

In particular, the two datasets I am using (used also in the CLIP paper) could be combined in one single benchmark in which we:

- train on ImageNet (linear evaluation and fine tuning both available)
- evaluate on ImageNet test or validation set
- evaluate on ImageNet-A to assess robustness against adversarial example
- evaluate on ImageNet-R to assess robustness against rendition shifts

Indeed, there should be no need to run the training several time: we should be able to report the accuracy on all 3 (imagenet, imagenet-a, imagenet-r) and possibly more.

I see two main ways to deal with this (not necessarily exclusive):

1. we create a benchmark that simply checks the performance on a fully trained network on imagenet (I think I saw an option to evaluate the head as well as the trunk)
2. we allow for more that one TEST set and several separate accuracy metrics. I do not think this is doable right now, and I am not sure it is worth it (your feedback is really important), although it could be used in other use cases as well (like HP tuning in VTAB equivalent, we could discuss that offline)


prigoyal (commented):
Thank you Quentin! Sharing thoughts below:
1. I think the name `robustness` is a bit generic. One could interpret all the benchmark suite as a means to test robustness. Do you have suggestion on other possible names?
2. I think a place to put the robustness could be `configs/config/robustness`  but depends on 1) above on alternate name suggestions.

extra_scripts/create_imagenet_r_data_files.py
```diff
@@ -0,0 +1,276 @@
...
+    URL = "https://people.eecs.berkeley.edu/~hendrycks/imagenet-r.tar"
+    download_and_extract_archive(url=URL, download_root=root)
+
+
+def import_imagenet_a_module():
+    script_path = os.path.split(__file__)[0]
+    return importlib.import_module("create_imagenet_a_data_files", script_path)
+
+
+IMAGENET_R_TARGET_MAPPINGS = {
```
prigoyal, line 52: Is there a way for us to upload this to AWS and retrieve the mapping from URL
```
module load  fairusers_aws

fs3cmd put my_file.txt s3://dl.fbaipublicfiles.com/vissl/data/my_file.txt
```

extra_scripts/create_imagenet_a_data_files.py
```diff
@@ -0,0 +1,298 @@
...
+
+    # Save the these lists in the disk_filelist format
+    os.makedirs(output_path, exist_ok=True)
+    img_info_out_path = os.path.join(output_path, "test_images.npy")
+    label_info_out_path = os.path.join(output_path, "test_labels.npy")
+    np.save(img_info_out_path, np.array(image_paths))
+    np.save(label_info_out_path, np.array(image_labels))
+
+
+IMAGENET_A_TARGET_MAPPINGS = {
```
prigoyal, line 76: Is there a way for us to upload this to AWS and retrieve the mapping from URL
```
module load  fairusers_aws

fs3cmd put my_file.txt s3://dl.fbaipublicfiles.com/vissl/data/my_file.txt
```

extra_scripts/create_imagenet_r_data_files.py
```diff
@@ -0,0 +1,276 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
```
prigoyal, line 2: Is it possible to add a comment explaining what this script is for here?

extra_scripts/create_imagenet_a_data_files.py
```diff
@@ -0,0 +1,298 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
```
prigoyal, line 2: Is it possible to add a comment explaining what this script is for here?

extra_scripts/create_imagenet_a_data_files.py
```diff
@@ -0,0 +1,298 @@
...
+    """
+    Create a test split for ImageNet, by reading the image folder in the
+    'input_path' and mapping its targets to the targets of imagenet
+    using the 'target_mappings' dictionary
+    """
+
+    # Map the images of the image folder to their corresponding targets
+    # in ImageNet
+    image_paths = []
+    image_labels = []
```
prigoyal, line 62: nit:
`image_paths, image_labels = [], []`

extra_scripts/create_imagenet_a_data_files.py
```diff
@@ -0,0 +1,298 @@
...
+    imagenet_r = datasets.ImageFolder(root=input_path, loader=lambda x: x)
+    for image_path, target_id in tqdm(imagenet_r):
+        image_paths.append(image_path)
+        image_labels.append(target_mappings[target_id])
+
+    # Save the these lists in the disk_filelist format
```
prigoyal, line 68: if the data is usable as a disk_folder, why do we need to convert it to disk_filelist? just curious :) I am sure you have a good reason :)

extra_scripts/create_imagenet_a_data_files.py
```diff
@@ -0,0 +1,298 @@
...
+    os.makedirs(output_path, exist_ok=True)
+    img_info_out_path = os.path.join(output_path, "test_images.npy")
+    label_info_out_path = os.path.join(output_path, "test_labels.npy")
+    np.save(img_info_out_path, np.array(image_paths))
+    np.save(label_info_out_path, np.array(image_labels))
```
prigoyal, line 73: Can you look into using the `PathManager` here? https://github.com/facebookresearch/vissl/blob/master/extra_scripts/create_voc_data_files.py#L16

extra_scripts/create_imagenet_r_data_files.py
```diff
@@ -0,0 +1,276 @@
...
+    Download the Imagenet-R dataset archive and expand it in the folder
+    provided as parameter
+    """
+    URL = "https://people.eecs.berkeley.edu/~hendrycks/imagenet-r.tar"
+    download_and_extract_archive(url=URL, download_root=root)
+
+
+def import_imagenet_a_module():
+    script_path = os.path.split(__file__)[0]
+    return importlib.import_module("create_imagenet_a_data_files", script_path)
```
prigoyal, line 49: I am not fully sure about this. Are we saying that we have a dependency on the other script ? Is it possible to expand the other script instead in that case and achieve `a` or `r` based on the user input?

docs/source/evaluations/robustness_benchmark.rst
```diff
@@ -0,0 +1,60 @@
...
+One particularly important type of benchmark is robustness benchmarks.
+Robustness benchmarks tell us how well a train model generalises on samples outside of its training distribution.
+
+These benchmarks are also interesting on models pre-trained using Self-Supervised Learning and have been used in the past in papers such as `CLIP <https://cdn.openai.com/papers/Learning_Transferable_Visual_Models_From_Natural_Language_Supervision.pdf>`_:
+they give a measure of how well an SSL algorithm is able to produce representations free of spurious correlations.
+
+VISSL provides benchmarks for Out-Of-Distribution (OOD) based on the two datasets `Imagenet-A <https://github.com/hendrycks/natural-adv-examples>`_ and `Imagenet-R <https://github.com/hendrycks/imagenet-r>`_.
+In those benchmarks, a pre-trained model is either fine-tuned or trained with linear evaluation on Imagenet, and then tested against the `Imagenet-A <https://github.com/hendrycks/natural-adv-examples>`_ and `Imagenet-R <https://github.com/hendrycks/imagenet-r>`_.
+
+For reproducibility, see `VISSL Model Zoo <https://github.com/facebookresearch/vissl/blob/master/MODEL_ZOO.md>`_.
```
prigoyal, line 15: is this relevant?

extra_scripts/create_imagenet_r_data_files.py
QuentinDuval: Yes, that's actually something I wanted to ask you: how to handle properly the resource files 👍 

I like the idea of moving that to AWS ! Thank you Priya :)

extra_scripts/create_imagenet_r_data_files.py
QuentinDuval: Sure ! I will do this :)

extra_scripts/create_imagenet_a_data_files.py
QuentinDuval: Yes, I am not 100% confident about this either.

Here is the key problem I tried to solve doing this: the benchmark consists in training Imagenet and then evaluating on Imagenet-A or Imagenet-R.

But Imagenet-A and Imagenet-R do not have the same amount of classes as Imagenet, and have instead a sub-set of the classes (200 classes among the 1000 classes). In addition to this, the classes do not match the first 200, but are a random sample of the whole 1000 classes.

=> So, if I load the dataset with ImageFolder, I end up with 200 targets, and these targets will be in the range [0, 200) and will not match their corresponding Imagenet targets in the range [0, 1000).

So I thought of 2 options:
1. add empty folders to the Imagenet-A and Imagenet-R datasets, these folders would make sure that the same number of folders (with the same name) appear in both Imagenet and Imagenet-A/R
2. use the disk_filelist format to remap the indices of Imagenet-A and Imagenet-R to their corresponding targets in Imagenet

I found the option 2 more elegant and went for it, but elegance is not a very tangible measure, I fully concede that :)

extra_scripts/create_imagenet_r_data_files.py
QuentinDuval: Yes, I can expand the other script instead.

I was merely trying to make sure that all scripts have the same interface (same options), but this clearly has a cost in terms of dependencies.

One option to keep the same uniform interface is to prepare both these datasets at the same time (and rename the script `create_imagenet_ood_test_splits.py`). What do you think?

docs/source/evaluations/robustness_benchmark.rst
QuentinDuval: You mean, the part about reproducibility?

growlix:
Just dropping in to mention that if you're benchmarking corruption-robustness, I would really encourage you to also use [ImageNet-C](https://github.com/hendrycks/robustness) :)

docs/source/evaluations/robustness_benchmark.rst
prigoyal: yep, I was wondering if we need to reference to model_zoo here :) sorry for being vague earlier :)

extra_scripts/create_imagenet_a_data_files.py
prigoyal: ah I see your point. I think given you care about the subsets, it makes sense to me to try to create the `disk_filelist`. Let's do this. :)

extra_scripts/create_imagenet_r_data_files.py
prigoyal: yes I think preparing at the same time and renaming as you suggested is great :)

prigoyal:
> Just dropping in to mention that if you're benchmarking corruption-robustness, I would really encourage you to also use [ImageNet-C](https://github.com/hendrycks/robustness) :)

a tangential thought but inspired by @growlix's comment above: a name like `corruption_robustness` could be a better folder name for us and it could be kept in `configs/config/benchmark/corruption_robustness` :) just one thought.

QuentinDuval:
> Just dropping in to mention that if you're benchmarking corruption-robustness, I would really encourage you to also use [ImageNet-C](https://github.com/hendrycks/robustness) :)

Yes, it's a good idea :)

I actually considered it, but since I was not able to find a reference (yet) of a SSL method using it (for instance, CLIP uses Imagenet-A and Imagenet-R, but not Imagenet-C), I decided to drop it.

But I think you're right, and it would be interesting nevertheless 👍

QuentinDuval:
> > Just dropping in to mention that if you're benchmarking corruption-robustness, I would really encourage you to also use [ImageNet-C](https://github.com/hendrycks/robustness) :)
> 
> a tangential thought but inspired by @growlix's comment above: a name like `corruption_robustness` could be a better folder name for us and it could be kept in `configs/config/benchmark/corruption_robustness` :) just one thought.

The issue is that Imagenet-R does not really fit the "corruption" term (and neither does Imagenet-A). But we could rename it like "ood_robustness" or "out_of_distribution" or a combination of both. The goal is to see whether the representation learned are robust to distribution shifts due to:
- selected example that are typically problematic (Imagenet-A)
- examples of the same classes but with different renditions (Imagenet-R)
- examples of images with blur / noise (Imagenet-C)

What do you think?

prigoyal:
> > > Just dropping in to mention that if you're benchmarking corruption-robustness, I would really encourage you to also use [ImageNet-C](https://github.com/hendrycks/robustness) :)
> > 
> > 
> > a tangential thought but inspired by @growlix's comment above: a name like `corruption_robustness` could be a better folder name for us and it could be kept in `configs/config/benchmark/corruption_robustness` :) just one thought.
> 
> The issue is that Imagenet-R does not really fit the "corruption" term (and neither does Imagenet-A). But we could rename it like "ood_robustness" or "out_of_distribution" or a combination of both. The goal is to see whether the representation learned are robust to distribution shifts due to:
> 
> * selected example that are typically problematic (Imagenet-A)
> * examples of the same classes but with different renditions (Imagenet-R)
> * examples of images with blur / noise (Imagenet-C)
> 
> @prigoyal, @growlix What do you think?

agreed! `robustness_out_of_distribution` sounds like a great name to me based on the above information :)

Issue #233, opened 2021-03-14
Wrong package apex gets installed while following the documentation


rhomu:
## 📚 VISSL Documentation

When installing VISSL using pip in a venv, the documentation states:
```
pip install apex -f https://dl.fbaipublicfiles.com/vissl/packaging/apexwheels/py38_cu101_pyt151/download.html
```
which installs the pypi package apex https://pypi.org/project/apex/. Instead the order of the arguments should be reversed:
```
pip install -f https://dl.fbaipublicfiles.com/vissl/packaging/apexwheels/py38_cu101_pyt151/download.html apex
```

System: Ubuntu 20.04.2 LTS, pip 20.0.2

prigoyal:
thank you for pointing this out. Assigning to @bottler to take a look.

PR #237, opened 2021-03-15
move apex after -f

bottler:
As noted in #233, because our apex packages are called `apex`, which clashes with a name on PyPI, we need to put the apex after the `-f <url>` on the pip command line.


prigoyal:
thank you @bottler , can you Import this to phabricator and let's merge it :)

PR #238, opened 2021-03-15
[feature] (VTAB) Low shot fine tuning configuration

QuentinDuval:
A first list of configurations for low shot (1000 samples) fine-tuning on VTAB datasets: all configurations have been tested on a SLURM cluster and ran to completion.

This PR depends on the DATA_LIMIT improvements of PR https://github.com/facebookresearch/vissl/pull/216 for the configuration to be valid.


prigoyal:
thank you @QuentinDuval , looks like the branch is based off of DATA_LIMIT change branch. Can you look into it ? :) ideally this branch can be independent of the other branches. :)

prigoyal:
thank you @QuentinDuval . I am wondering if we can instead keep one folder `low_shot_classification` and move the new sub-folders corresponding to each dataset name there?

QuentinDuval:
> thank you @QuentinDuval . I am wondering if we can instead keep one folder `low_shot_classification` and move the new sub-folders corresponding to each dataset name there?

Indeed, I can move them in `low_shot_classification` as it makes complete sense (the main difference with the benchmarks already there is that it is fine tuning and not linear probing).

The other option I had was to move them with the "fulltune" (but then I should rename it "fine_tune").

The last option I had was to create a folder named "vtab" since these benchmarks are very close to the VTAB protocol (we still miss the search for hyper-parameters, implemented the way it is done in VTAB, to make it VTAB though).

I was not 100% sure how to proceed.

@prigoyal What do you think would be the best option?

prigoyal:
I think we should keep it simplistic and simply move under `low_shot_classification`. This folder contains not just linear probe (for example SVM trainings in case of voc07). What we can do is to rename `low_shot_classification` to `low_shot_transfer` to make it generic and then move everything under it. Thoughts?

QuentinDuval:
> I think we should keep it simplistic and simply move under `low_shot_classification`. This folder contains not just linear probe (for example SVM trainings in case of voc07). What we can do is to rename `low_shot_classification` to `low_shot_transfer` to make it generic and then move everything under it. Thoughts?

I think you are right: we should keep it simple 👍 

Let us go with `low_shot_transfer`, as you proposed, as it describes it all.

I will update the PR today to reflect these changes !

docs/source/evaluations/fine_tuning.rst
```diff
@@ -38,14 +41,178 @@ Benchmark: Places205
...
 
-
 .. code-block:: bash
 
     python tools/run_distributed_engines.py \
       config=benchmark/fulltune/places205/eval_resnet_8gpu_transfer_places205_fulltune \
       config.MODEL.WEIGHTS_INIT.PARAMS_FILE=<my_weights.torch>
 
 
+Benchmark: CIFAR-100
```
prigoyal, line 51: I think we can move the low-shot section out to a dedicated `low_shot_transfer.rst` to make it more visible :)

prigoyal (commented):
looks great! small inline comment + can you verify once we need to make changes to docs/readmes etc due to the renaming of `low_shot_classification` ? 

otherwise looking awesome and exciting! :)

docs/source/evaluations/fine_tuning.rst
QuentinDuval: > can you verify once we need to make changes to docs/readmes etc due to the renaming of `low_shot_classification` ?

Indeed, to my surprise, I did not see any reference to the `low_shot_classification` folder. I will create a dedicated documentation as you suggests, it would indeed make sense to separate it from the full fine-tuning, and we can then enrich it with the two others tasks (VOC07 and Places205).

QuentinDuval: I moved the documentation in a specific file in 27f53fe.

I propose we complete it with the VOC07 and Places205 in another PR (I first need to play with these configuration to better understand them): I can create an issue with it and add it to my backlog.

prigoyal: sounds great to me :)

PR #251, opened 2021-03-24
[Proposal] [Draft] [tooling] configurable linting script

QuentinDuval:
Configurable linting script which allows to select whether or not:
- to lint for the whole repo or just the modified files
- select which type of linting to apply (useful for fixing error and skipping to the selected check)
- with option to decide for auto-correction or not

The main disadvantage of the script is that it requires an additional dependency, reason why I did not replace "linting.sh" but the counterpart is much greater flexibility.

I let you decide if this is interesting: I use it on my side, so maybe it might be useful to others, but I can definitely keep it on my side only if necessary.

### DIFF
--- a/dev/README.md
+++ b/dev/README.md
@@ -21,7 +21,7 @@ For every PR, we run/mandate a few checks before code is ready for review. The c
 2. **black**: We format the code using black8
 3. **isort**: We check that imports are properly sorted https://pypi.org/project/isort/. See `setup.cfg` for the settings.
 
-In order to format code before code review, there are 2 options:
+In order to format code before code review, there are several options:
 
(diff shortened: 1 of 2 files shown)


prigoyal (commented):
I think this is a useful tool. Can we expand the README https://github.com/facebookresearch/vissl/tree/master/dev#option-1-use-devlintersh as well    :)

prigoyal:
let's also import it to phabricator afterwards and aim to land it :)

QuentinDuval:
> I think this is a useful tool. Can we expand the README https://github.com/facebookresearch/vissl/tree/master/dev#option-1-use-devlintersh as well :)

I added an option 3 in the commit b0606d5 :)

PR #252, opened 2021-03-24
[fix] If nvidia-smi returns with non zero code, log the error instead of interrupting the training

QuentinDuval:
I got this error during training (nvidia-smi returning a non-zero code) leading to the termination of my training.

If this happens, I think we should definitely log the event, but not interrupt the training.

Here is a short fix to deal with this issue.


prigoyal:
looks good. can you import it and let's land it :)

PR #256, opened 2021-03-25
[readme] (tutorial) Make GitHub repository the source of truth for tutorials

QuentinDuval:
A quick fix to move the source of truth for the tutorials to the GitHub repository: this way we avoid the sharing issues as well as we make sure that the content of the "tutorial" folder is in line with the links in the README.

Tell me what you think about it :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,12 +41,12 @@ Learn more about VISSL at our [documentation](https://vissl.readthedocs.io). And
 
 Get started with VISSL by trying one of the **Colab tutorial notebooks**.
 
-- [Train SimCLR on 1-gpu](https://colab.research.google.com/drive/1Rt3Plt3ph84i1A-eolLFafybwjrBFxYe?usp=sharing)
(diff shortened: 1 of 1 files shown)


prigoyal:
thank you! certainly an improvement and much welcome change. I think we should link to `stable` instead of `master` though so that when users use the tutorials and they install the pip packages , then they will have right code expectations corresponding to the stable tag. (simply replace master with stable in the links and also verify the stable links work) 
thoughts?

QuentinDuval:
> thank you! certainly an improvement and much welcome change. I think we should link to `stable` instead of `master` though so that when users use the tutorials and they install the pip packages , then they will have right code expectations corresponding to the stable tag. (simply replace master with stable in the links and also verify the stable links work)
> thoughts?

I think you just point to a very important issue: we need to make sure that the version associated to a branch are the ones pointed by that branch:
- on the master, it should link to the master
- on the v0.1.5, it should link to this version

I need to think about how to do this.

QuentinDuval:
To be addressed later (we first need to give the rights to all tutorials as we discussed)

PR #264, opened 2021-03-30
[feature] (activation stats): ability to track the statistics of each layer in the model when training

QuentinDuval:
# Tracking activation statistics during training

This PR features a new monitoring utility, plugged in VISSL via the Tensorboard hook, which allows to capture the output of the "leaf modules" (not all modules) and compute mean and spread statistics on the output of each layer, and track them in tensorboard.

- [ ] @QuentinDuval to re-run performance tests after review comments and refactoring on SwAV
- [ ] @QuentinDuval to try to add some quick performance tests in the unit tests

## Output example:

Screenshot 2021-04-07 at 08 52 10

## Description

The configuration has been updated with the following option (by default disabled):

```
  MONITORING:
    # At which frequency do we monitor statistics on the activations:
    # - 0 means that we do not monitor statistics
    # - N > 0 means we monitor every N iterations
    MONITOR_ACTIVATION_STATISTICS: 0
```

Turn on the option with the following hydra override: `config.MONITORING.MONITOR_ACTIVATION_STATISTICS=50` to gather the statistics on all activations every 50 iterations.

**NOTE:** This option requires the tensorboard hook to be enabled to take effect.

**Performance impacts:** The collect of statistics makes use of the following optimisation: for feature maps, we only compute the statistics on the central feature, which requires less compute and memory than on the full feature map and still exercises all the weights of the BN or Conv2D layer. With this optimisation, the impact in terms of memory is negligible and the impact in terms of runtime is about 1.4% on SimCLR, when the frequency of collection is every 50 iterations.

## Further improvements

- Support for other backend than just Tensorboard: ideally, we should be able to dump raw data / images
- Profile what takes time when computing statistics and optimise it further if possible
- Adding an "alerting" feature which varies the speed of tracking based on detected divergence

### DIFF
--- a/tests/test_activation_statistics.py
+++ b/tests/test_activation_statistics.py
@@ -1,0 +1,48 @@
+import unittest
+
+import torch
+import torchvision.models as models
+from vissl.utils.activation_statistics import (
+    ActivationStatisticsAccumulator,
+    ActivationStatisticsMonitor,
+)
+
+
+class TestDataLimitSubSampling(unittest.TestCase):
+    def test_activation_statistics(self):
+        torch.manual_seed(0)
+
+        accumulator = ActivationStatisticsAccumulator()
+        watcher = ActivationStatisticsMonitor(
+            observer=accumulator, log_frequency=1, ignored_modules=set()
+        )
+        watcher.set_iteration(1)
+        model = models.resnet18()
+        watcher.monitor(model)
+        model(torch.randn(size=(1, 3, 224, 224)))
+
+        stats = accumulator.stats
+        self.assertEqual(60, len(stats))
+        for stat in stats:
+            self.assertEqual(1, stat.iteration)
+
+        # Verify that the first statistics produced is for the first
+        # layer of the ResNet
+        first_stat = stats[0]
+        self.assertEqual("conv1", first_stat.name)
+        self.assertEqual("torch.nn.modules.conv.Conv2d", first_stat.module_type)
+        self.assertAlmostEqual(-0.0001686, first_stat.mean, delta=1e-6)
+        self.assertAlmostEqual(1.6035640, first_stat.maxi, delta=1e-6)
+        self.assertAlmostEqual(-1.4944878, first_stat.mini, delta=1e-6)
+
+        # Verify that only leaf modules have statistics
+        exported_modules_types = {
+            "torch.nn.modules.pooling.MaxPool2d",
+            "torch.nn.modules.pooling.AdaptiveAvgPool2d",
+            "torch.nn.modules.batchnorm.BatchNorm2d",
+            "torch.nn.modules.activation.ReLU",
+            "torch.nn.modules.conv.Conv2d",
+            "torch.nn.modules.linear.Linear",
+        }
+        module_types = {stat.module_type for stat in stats}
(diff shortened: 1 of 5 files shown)


prigoyal:
Hi @QuentinDuval , all the points you raised are great. I'd propose that we meet over VC to discuss these.

certainly interested in the fixes, feel free to make PR :)

prigoyal (commented):
I did a high level design overview of this and it looks great to me! :) early feedback but it is heading in the right direction :) great work @QuentinDuval :)

tests/test_activation_statistics.py
```diff
@@ -0,0 +1,82 @@
+import unittest
```
prigoyal, line 1: nit: we should add the Copyright header to all python modules. See other python modules :)

vissl/utils/activation_statistics.py
```diff
@@ -0,0 +1,181 @@
...
+        """
+        Install hooks on the model to track its activation statistics
+        """
+        for name, m in model.named_modules():
+            if self._get_qualified_type(m) not in self.ignored_modules:
+                h1 = m.register_forward_pre_hook(self._create_pre_forward_hook(name))
+                h2 = m.register_forward_hook(self._create_post_forward_hook(name))
+                self._hooks.extend([h1, h2])
+
+    def stop(self):
```
prigoyal, line 94: nice :)

prigoyal (commented):
looks great to me @QuentinDuval , no comments code wise. Next steps as we discussed , and then it's good to go :)

prigoyal (commented):
this is super awesome work @QuentinDuval :)

quick clarification: leaf modules -> doesn't mean just head is tracked right? from the code, it seems like trunk+head both are tracked but figures in test plan are for the heads so just wanted to double check :)

QuentinDuval:
> this is super awesome work @QuentinDuval :)
> 
> 1. quick clarification: leaf modules -> doesn't mean just head is tracked right? from the code, it seems like trunk+head both are tracked but figures in test plan are for the heads so just wanted to double check :)
> 2. we should insert the copyright headers everywhere -> blocker for landing this PR

Hi @prigoyal :)

By leaf module, I meant modules that are not encapsulating other modules. For instance, `nn.Sequential(nn.Linear(...), nn.ReLU(...))`: the leaf modules are nn.Linear() and nn.ReLU() and we ignore nn.Sequential. This also means we will ignore things like Bottleneck (which only contains modules) but might also ignore some interesting modules now that I think about it (*).

Indeed, otherwise, all modules are being monitored if they are in training mode: we ignore modules that are not trained, like frozen modules when doing a linear evaluation, for there is not much to monitor in that case.

(*) I think we can improve that in later PR: ignore modules who do not have parameters on their own but only parameters in their children, or simply hardcode what we want to ignore.

QuentinDuval:
> this is super awesome work @QuentinDuval :)
> 
> 1. quick clarification: leaf modules -> doesn't mean just head is tracked right? from the code, it seems like trunk+head both are tracked but figures in test plan are for the heads so just wanted to double check :)
> 2. we should insert the copyright headers everywhere -> blocker for landing this PR

I added the copyrights in d85434b 👍  Good catch !

PR #265, opened 2021-03-31
[benchmark] (SUN397) Add SUN397 dataset and associated benchmarks

QuentinDuval:
This PR adds the SUN397 dataset using in VTAB, with automatic download supported, with the splits created similarly to what is being done in VTAB, and includes the configuration and documentation for the associated benchmark:
- one for linear evaluation on the full dataset (around 87000 images)
- one for VTAB-like low shot transfer on 1000 images (challenging benchmark since there are 397 classes)

I tested the performance on SimCLR and the results are in line with was is reported in the SimCLR paper.

One more step in the direction of VTAB / CLIP like benchmark suites :)

### DIFF
--- a/extra_scripts/create_sun397_data_files.py
+++ b/extra_scripts/create_sun397_data_files.py
@@ -4,11 +4,22 @@ import argparse
 import dataclasses
 import os
 import random
-from typing import List
+from typing import Any, List
 
 import numpy as np
+from fvcore.common.file_io import PathManager
 from torchvision.datasets.utils import download_and_extract_archive
 from tqdm import tqdm
+
+
+@dataclasses.dataclass
+class SplitData:
+    """
+    Data structure holding the samples associated to a given split
+    """
+
+    image_paths: List[str] = dataclasses.field(default_factory=list)
+    image_labels: List[int] = dataclasses.field(default_factory=list)
 
 
 def get_argument_parser():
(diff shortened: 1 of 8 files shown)


extra_scripts/create_sun397_data_files.py
```diff
@@ -0,0 +1,146 @@
...
+    URL = "http://vision.princeton.edu/projects/2010/SUN/SUN397.tar.gz"
+    download_and_extract_archive(url=URL, download_root=root)
+
+
+TRAIN_SPLIT_SIZE = 0.7
+VALID_SPLIT_SIZE = 0.1
+
+
+@dataclasses.dataclass
+class SplitData:
```
prigoyal, line 62: nit: let's move this right after the import statements :)

extra_scripts/create_sun397_data_files.py
```diff
@@ -0,0 +1,146 @@
...
+    """
+    Data structure holding the samples associated to a given split
+    """
+
+    image_paths: List[str] = dataclasses.field(default_factory=list)
+    image_labels: List[int] = dataclasses.field(default_factory=list)
+
+
+def split_sample_list(xs):
```
prigoyal, line 71: nit: use a more descriptive input name instead of `xs` :)

extra_scripts/create_sun397_data_files.py
```diff
@@ -0,0 +1,146 @@
...
+                image_path = os.path.join(full_class_path, image_name)
+                splits_data[split].image_paths.append(image_path)
+                splits_data[split].image_labels.append(i)
+
+    # Save each split
+    for split, samples in splits_data.items():
+        image_output_path = os.path.join(output_path, f"{split}_images.npy")
+        label_output_path = os.path.join(output_path, f"{split}_labels.npy")
+        np.save(image_output_path, np.array(samples.image_paths))
+        np.save(label_output_path, np.array(samples.image_labels))
```
prigoyal, line 130: we should use the `PathManager` :) it will ensure that we can use any file storage system and not just disk :)

prigoyal (commented):
looks awesome @QuentinDuval ! small inline comments but good to go otherwise :)

extra_scripts/create_sun397_data_files.py
QuentinDuval: Oups... this "xs" came right out of my Haskell days ! I will rename this :)

extra_scripts/create_sun397_data_files.py
QuentinDuval: Fixed at a3f973b !

extra_scripts/create_sun397_data_files.py
QuentinDuval: Fixed at a3f973b :)

PR #271, opened 2021-04-06
working prototype of wandb

pclucas14:
This is a working prototype of a wandb logger. I'm filing the PR now so the team can have a look, but there are probably still a few things to iron out first.

### DIFF
--- a/vissl/hooks/wandb_hook.py
+++ b/vissl/hooks/wandb_hook.py
@@ -1,0 +1,226 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+import logging
+
+import torch
+from classy_vision import tasks
+from classy_vision.generic.distributed_util import is_primary
+from classy_vision.hooks.classy_hook import ClassyHook
+
+if is_primary():
+    import wandb
+
+BYTE_TO_MiB = 2 ** 20
+
+class SSLWandbHook(ClassyHook):
+
+    on_loss_and_meter = ClassyHook._noop
+    on_backward = ClassyHook._noop
+    on_start = ClassyHook._noop
+    on_end = ClassyHook._noop
+    on_step = ClassyHook._noop
+
+    def __init__(
+        self,
+        log_params: bool = False,
+        log_params_every_n_iterations: int = -1,
+        log_params_gradients: bool = False,
+    ) -> None:
+        """The constructor method of SSLWandbHook.
+
+        Args:
+            log_params (bool): whether to log model params to tensorboard
+            log_params_every_n_iterations (int): frequency at which parameters
+                        should be logged to tensorboard
+            log_params_gradients (bool): whether to log params gradients as well
+                        to tensorboard.
+        """
+        super().__init__()
+        # going to assume WandB install check is already performed (TODO: check this)
+
+        logging.info("Setting up SSL Wandb Hook...")
+        self.watched = False
+        self.log_params = log_params
+        self.log_params_every_n_iterations = log_params_every_n_iterations
+        self.log_params_gradients = log_params_gradients
+        logging.info(
+            f"Wandb config: log_params: {self.log_params}, "
+            f"log_params_freq: {self.log_params_every_n_iterations}, "
+            f"log_params_gradients: {self.log_params_gradients}"
+        )
+
+
+    def on_forward(self, task: "tasks.ClassyTask") -> None:
+        """
+        Called after every forward if tensorboard hook is enabled.
+        Logs the model parameters if the training iteration matches the
+        logging frequency.
+        """
+        if not self.log_params:
+            return
+
+        if (
+            self.log_params_every_n_iterations > 0
+            and is_primary()
+            and task.train
+            and task.iteration % self.log_params_every_n_iterations == 0
+        ):
+            out_dict = {}
+            for name, parameter in task.base_model.named_parameters():
+                parameter = parameter.cpu().data.numpy()
+                out_dict[f"Parameters/{name}"] = wandb.Histogram(parameter)
+
+            wandb.log(out_dict, step=task.iteration)
+
+    def on_phase_start(self, task: "tasks.ClassyTask") -> None:
+        """
+        Called at the start of every epoch if the tensorboard hook is
+        enabled.
+        Logs the model parameters once at the beginning of training only.
+        """
+        if not self.log_params:
+            return
+
+        # log the parameters just once, before training starts
+        if is_primary() and task.train and task.train_phase_idx == 0:
+            out_dict = {}
+            for name, parameter in task.base_model.named_parameters():
+                parameter = parameter.cpu().data.numpy()
+                out_dict[f"Parameters/{name}"] = wandb.Histogram(parameter)
+
+            wandb.log(out_dict, step=task.iteration)
+
+    def on_phase_end(self, task: "tasks.ClassyTask") -> None:
+        """
+        Called at the end of every epoch if the tensorboard hook is
+        enabled.
+        Log model parameters and/or parameter gradients as set by user
+        in the tensorboard configuration. Also resents the CUDA memory counter.
+        """
+        out_dict = {}
+
+        # Log train/test accuracy
+        if is_primary():
+            phase_type = "Training" if task.train else "Testing"
+            for meter in task.meters:
+                if "accuracy" in meter.name:
+                    for top_n, accuracies in meter.value.items():
+                        for i, acc in accuracies.items():
+                            tag_name = f"{phase_type}/Accuracy_" f" {top_n}_Output_{i}"
+                            out_name[tag_name] = round(acc, 5)
+
+        if not (self.log_params or self.log_params_gradients):
+            if len(out_dict) > 0:
+                wandb.log(out_dict, step=task.iteration)
+            return
+
+        if is_primary() and task.train:
+            # Log the weights and bias at the end of the epoch
+            if self.log_params:
+                for name, parameter in task.base_model.named_parameters():
+                    parameter = parameter.cpu().data.numpy()
+                    out_dict[f"Parameters/{name}"] = wandb.Histogram(parameter)
+
+            # Log the parameter gradients at the end of the epoch
+            if self.log_params_gradients:
+                for name, parameter in task.base_model.named_parameters():
+                    if parameter.grad is not None:
+                        try:
+                            parameter = parameter.grad.cpu().data.numpy()
+                            out_dict[f"Gradients/{name}"] = wandb.Histogram(parameter)
+                        except ValueError:
+                            logging.info(
+                                f"Gradient histogram empty for {name}, "
+                                f"iteration {task.iteration}. Unable to "
+                                f"log gradient."
+                            )
+
+            # Reset the GPU Memory counter
+            if torch.cuda.is_available():
+                torch.cuda.reset_max_memory_allocated()
+                torch.cuda.reset_max_memory_cached()
+
+            wandb.log(out_dict, step=task.iteration)
+
+
+    def on_update(self, task: "tasks.ClassyTask") -> None:
+        """
+        Called after every parameters update if tensorboard hook is enabled.
+        Logs the parameter gradients if they are being set to log,
+        log the scalars like training loss, learning rate, average training
+        iteration time, batch size per gpu, img/sec/gpu, ETA, gpu memory used,
+        peak gpu memory used.
+        """
+
+        if not is_primary():
+            return
+
+        out_dict = {}
+        iteration = task.iteration
+
+        if (
+            self.log_params_every_n_iterations > 0
+            and self.log_params_gradients
+            and task.train
+            and iteration % self.log_params_every_n_iterations == 0
+        ):
+            logging.info(f"Logging Parameter gradients. Iteration {iteration}")
+            for name, parameter in task.base_model.named_parameters():
+                if parameter.grad is not None:
+                    try:
+                        parameter = parameter.grad.cpu().data.numpy()
+                        out_dict[f"Gradients/{name}"] = wandb.Histogram(parameter)
+                    except ValueError:
+                        logging.info(
+                            f"Gradient histogram empty for {name}, "
+                            f"iteration {task.iteration}. Unable to "
+                            f"log gradient."
+                        )
+
+        if iteration % task.config["LOG_FREQUENCY"] == 0 or (
+            iteration <= 100 and iteration % 5 == 0
+        ):
+            logging.info(f"Logging metrics. Iteration {iteration}")
+            out_dict["Training/Loss"] = round(task.last_batch.loss.data.cpu().item(), 5)
+            out_dict["Training/Learning_rate"] = round(task.optimizer.options_view.lr, 5)
+
+            # Batch processing time
+            if len(task.batch_time) > 0:
+                batch_times = task.batch_time
+            else:
+                batch_times = [0]
+
+            batch_time_avg_s = sum(batch_times) / max(len(batch_times), 1)
+            out_dict["Speed/Batch_processing_time_ms"] = scalar_value=int(1000.0 * batch_time_avg_s)
+
+            # Images per second per replica
+            pic_per_batch_per_gpu = task.config["DATA"]["TRAIN"][
+                "BATCHSIZE_PER_REPLICA"
+            ]
+            pic_per_batch_per_gpu_per_sec = (
+                int(pic_per_batch_per_gpu / batch_time_avg_s)
+                if batch_time_avg_s > 0
+                else 0.0
+            )
+            out_dict["Speed/img_per_sec_per_gpu"] = pic_per_batch_per_gpu_per_sec
+
+            # ETA
+            avg_time = sum(batch_times) / len(batch_times)
+            eta_secs = avg_time * (task.max_iteration - iteration)
+            out_dict["Speed/ETA_hours"] = eta_secs / 3600.0
+
+            # GPU Memory
+            if torch.cuda.is_available():
+                # Memory actually being used
+                out_dict["Memory/Peak_GPU_Memory_allocated_MiB"] = \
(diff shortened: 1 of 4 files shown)


prigoyal:
also cc @QuentinDuval and @min-xu-ai :)

pclucas14:
Right now a critical issue is that for some errors, wandb stalls instead of crashing. When testing this, if a run stalls, rerun it with wandb disabled (by running `wandb disabled` in the command line)

min-xu-ai:
> Right now a critical issue is that for some errors, wandb stalls instead of crashing. When testing this, if a run stalls, rerun it with wandb disabled (by running `wandb disabled` in the command line)

This is exciting. Can you share a bit more info on how to test this out? I can check out your branch and I suppose I need to install some wandb pip packages and etc.? Can you share the commands for folks to test it out and to see how it is used in action?

pclucas14:
Hi, 

to install wandb `pip install wandb` should do it. Some basic stuff you should know : There are essentially three `wandb` modes. 

1. `wandb on` will enable wandb and will sync your runs directly to the cloud, as well as log files internally
2. `wandb off` does not upload online, only saves files locally, which you can upload later 
3. `wandb disabled` makes all wandb operations no-ops and is equivalent to not doing anything. 

If you don't have a wandb account you should start by creating one. Once that's done, wandb will give you instructions on what to set up on your remote server. 

In terms of reproducing the error, to me if you do distributed training (num_nodes > 1) and put a super large batch size leading to OOM, wandb currently stalls and I'm not sure why. It may also be an issue specific to me, I will also investigate.

pclucas14:
This is the exact command I am running : 
```
PYTHONPATH=./ python tools/run_distributed_engines.py config=pretrain/simclr/simclr_wandb.yaml config.DISTRIBUTED.NUM_PROC_PER_NODE=2 config.HOOKS.WANDB_SETUP.USE_WANDB=True config.DISTRIBUTED.NUM_NODES=1 config.HOOKS.WANDB_SETUP.EXP_NAME=simclr_mini_2gpu config.DATA.TRAIN.BATCHSIZE_PER_REPLICA=1000
```
running after `wandb disabled` gives crashes giving an OOM error, as seen here : 

```
INFO 2021-04-06 11:45:33,467 trainer_main.py: 301: Phase advanced. Rank: 1
INFO 2021-04-06 11:45:33,467 state_update_hooks.py:  98: Starting phase 16 [train]
INFO 2021-04-06 11:45:34,856 trainer_main.py: 301: Phase advanced. Rank: 0
INFO 2021-04-06 11:45:34,857 state_update_hooks.py:  98: Starting phase 16 [train]
Traceback (most recent call last):
  File "tools/run_distributed_engines.py", line 58, in <module>
    hydra_main(overrides=overrides)
  File "tools/run_distributed_engines.py", line 45, in hydra_main
    hook_generator=default_hook_generator,
  File "/private/home/lucaspc/repos/vissl/vissl/utils/distributed_launcher.py", line 145, in launch_distributed
    daemon=False,
  File "/private/home/lucaspc/.conda/envs/vissl2/lib/python3.7/site-packages/torch/multiprocessing/spawn.py", line 199, in spawn
    return start_processes(fn, args, nprocs, join, daemon, start_method='spawn')
  File "/private/home/lucaspc/.conda/envs/vissl2/lib/python3.7/site-packages/torch/multiprocessing/spawn.py", line 157, in start_processes
    while not context.join():
  File "/private/home/lucaspc/.conda/envs/vissl2/lib/python3.7/site-packages/torch/multiprocessing/spawn.py", line 118, in join
    raise Exception(msg)
Exception:

....

    self.padding, self.dilation, self.groups)
RuntimeError: CUDA out of memory. Tried to allocate 1.10 GiB (GPU 1; 15.78 GiB total capacity; 12.60 GiB already allocated; 641.75 MiB free; 13.93 GiB reserved in total by PyTorch)


```


With `wandb on` the method does not crash, but rather stalls. This is the last thing shown in the prompt:
```
INFO 2021-04-06 11:41:48,606 trainer_main.py: 301: Phase advanced. Rank: 1
INFO 2021-04-06 11:41:48,606 state_update_hooks.py:  98: Starting phase 16 [train]
INFO 2021-04-06 11:41:49,963 trainer_main.py: 301: Phase advanced. Rank: 0
INFO 2021-04-06 11:41:49,964 state_update_hooks.py:  98: Starting phase 16 [train]
```

vissl/utils/wandb.py
```diff
@@ -0,0 +1,128 @@
...
+    Returns:
+        wandb_available (bool): based on wandb imports, returns whether tensboarboard
+                             is available or not.
+    """
+    try:
+        import wandb  # noqa F401
+
+        wandb_available = True
+    except ImportError:
+        logging.info("Tensorboard is not available")
```
min-xu-ai, line 28: stale msg

vissl/utils/wandb.py
```diff
@@ -0,0 +1,128 @@
...
+
+    Args:
+        cfg (AttrDict): User specified config file containing the settings for the
+                        wandb as well like log directory, logging frequency etc
+
+    Returns:
+        SSLTensorboardHook (function): the wandb hook constructed
+    """
+    import wandb
+    from vissl.hooks import SSLWandbHook
```
min-xu-ai, line 82: move these to the top or is it due to circular imports?

vissl/utils/wandb.py
```diff
@@ -0,0 +1,128 @@
...
+            for key in current:
+                recurse(key_array + [key])
+
+    recurse([])
+    return final
+
+
+def get_wandb_hook(cfg):
+    """
+    Construct the Tensorboard hook for visualization from the specified config
```
min-xu-ai, line 72: Tensorboard?

vissl/hooks/wandb_hook.py
```diff
@@ -0,0 +1,226 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+
+import logging
+
+import torch
+from classy_vision import tasks
+from classy_vision.generic.distributed_util import is_primary
+from classy_vision.hooks.classy_hook import ClassyHook
+
+if is_primary():
```
min-xu-ai, line 10: why do this check? Importing on all ranks should be fine too?

vissl/hooks/wandb_hook.py
```diff
@@ -0,0 +1,226 @@
...
+        logging.info(
+            f"Wandb config: log_params: {self.log_params}, "
+            f"log_params_freq: {self.log_params_every_n_iterations}, "
+            f"log_params_gradients: {self.log_params_gradients}"
+        )
+
+
+    def on_forward(self, task: "tasks.ClassyTask") -> None:
+        """
+        Called after every forward if tensorboard hook is enabled.
```
min-xu-ai, line 55: tensorboard

min-xu-ai (commented):
I don't see how this hook can cause the stall when OOM should have been raised. Maybe it wasn't really a stall? How about the case without an OOM, does the training go on as normal? Doesn't get slow down when wandb is enabled? I'd suggest put some prints in the hook and the train loop to see where exactly the stall happens if it indeed stalls.

vissl/hooks/wandb_hook.py
pclucas14: wandb import has some overhead to it, so I was trying to limit it to the main worker since it's the only one using it. This can be changed

vissl/utils/wandb.py
pclucas14: yes

vissl/utils/wandb.py
pclucas14: I want to only import wandb if the program is calling the hook. This way it will still work if a user has not installed wandb yet.

vissl/utils/wandb.py
pclucas14: yes. fixed

vissl/hooks/wandb_hook.py
pclucas14: yes. fixed

pclucas14:
> I don't see how this hook can cause the stall when OOM should have been raised. Maybe it wasn't really a stall? How about the case without an OOM, does the training go on as normal? Doesn't get slow down when wandb is enabled? I'd suggest put some prints in the hook and the train loop to see where exactly the stall happens if it indeed stalls.

Ok, I will try this tomorrow and let you know. Note that this is not specific to an OOM error, I saw the same thing with another error on the data side.

pclucas14:
Hi, quick update : 

I'm unable to use the current logger when launching jobs with SLURM. If I ask for an interactive session and launch a job it works ok, however launching directly with slurm I get the following errror : 

```
-- Process 3 terminated with the following error:
Traceback (most recent call last):
  File "/private/home/lucaspc/.conda/envs/vissl2/lib/python3.7/site-packages/torch/multiprocessing/spawn.py", line 19, in _wrap
    fn(i, *args)
  File "/private/home/lucaspc/repos/vissl/vissl/utils/distributed_launcher.py", line 200, in _distributed_worker
    process_main(cfg, dist_run_id, local_rank=local_rank, node_id=node_id)
  File "/private/home/lucaspc/repos/vissl/vissl/utils/distributed_launcher.py", line 193, in process_main
    hook_generator=hook_generator,
  File "/private/home/lucaspc/repos/vissl/vissl/engines/train.py", line 93, in train_main
    hooks = hook_generator(cfg)
  File "/private/home/lucaspc/repos/vissl/vissl/hooks/__init__.py", line 126, in default_hook_generator
    wandb_hook = get_wandb_hook(cfg)
  File "/private/home/lucaspc/repos/vissl/vissl/utils/wandb.py", line 120, in get_wandb_hook
    name=name
  File "/private/home/lucaspc/.conda/envs/vissl2/lib/python3.7/site-packages/wandb/sdk/wandb_init.py", line 742, in init
    run = wi.init()
  File "/private/home/lucaspc/.conda/envs/vissl2/lib/python3.7/site-packages/wandb/sdk/wandb_init.py", line 513, in init
    raise UsageError(error_message)
wandb.errors.UsageError: Error communicating with wandb process

srun: error: learnfair0928: task 0: Exited with exit code 1
srun: launch/slurm: _step_signal: Terminating StepId=39934341.0
```

vissl/hooks/wandb_hook.py
prigoyal: I think it makes sense to me to do it on primary_rank only. :)

vissl/utils/wandb.py
```diff
@@ -0,0 +1,128 @@
...
+import logging
+
+from classy_vision.generic.distributed_util import is_primary
+from vissl.utils.checkpoint import get_checkpoint_folder
+from vissl.utils.io import makedir
+from omegaconf import DictConfig
+
+def is_wandb_available():
+    """
+    Check whether wandb is available or not.
```
prigoyal, line 17: can we add a link here to wandb documentation that open source users can read/access? :)

vissl/utils/wandb.py
```diff
@@ -0,0 +1,128 @@
...
+
+    Returns:
+        wandb_available (bool): based on wandb imports, returns whether tensboarboard
```
prigoyal, line 20: nit: tensboarboard -> wandb

vissl/utils/wandb.py
prigoyal: makes sense to me to do local imports .

vissl/utils/wandb.py
```diff
@@ -0,0 +1,128 @@
...
+
+        if os.path.exists(wandb_id_save_path):
+            wandb_id = open(wandb_id_save_path, 'r').read().splitlines()[0]
+        else:
+            wandb_id = wandb.util.generate_id()
+            with open(wandb_id_save_path, 'w') as f:
+                f.write(wandb_id)
+
+        name = cfg.HOOKS.WANDB_SETUP.EXP_NAME
+        if name == "??":
```
prigoyal, line 111: nit: let's set name = "" instead of ??

vissl/hooks/wandb_hook.py
```diff
@@ -0,0 +1,226 @@
...
+from classy_vision import tasks
+from classy_vision.generic.distributed_util import is_primary
+from classy_vision.hooks.classy_hook import ClassyHook
+
+if is_primary():
+    import wandb
+
+BYTE_TO_MiB = 2 ** 20
+
+class SSLWandbHook(ClassyHook):
```
prigoyal, line 15: A question on this: is this hook similar to Tensorboard hook with the only difference being in logging to "wandb" instead of tensorboard? 

Is it possible that we can inherit the `TensorboardHook` ? Or alternatively, does it make sense to extend the `TensorboardHook` directly to optionally log to wandb as well is user is using WandB ?

prigoyal (commented):
thank you @pclucas14 :)

prigoyal:
> Hi, quick update :
> 
> I'm unable to use the current logger when launching jobs with SLURM. If I ask for an interactive session and launch a job it works ok, however launching directly with slurm I get the following errror :
> 
> ```
> -- Process 3 terminated with the following error:
> Traceback (most recent call last):
>   File "/private/home/lucaspc/.conda/envs/vissl2/lib/python3.7/site-packages/torch/multiprocessing/spawn.py", line 19, in _wrap
>     fn(i, *args)
>   File "/private/home/lucaspc/repos/vissl/vissl/utils/distributed_launcher.py", line 200, in _distributed_worker
>     process_main(cfg, dist_run_id, local_rank=local_rank, node_id=node_id)
>   File "/private/home/lucaspc/repos/vissl/vissl/utils/distributed_launcher.py", line 193, in process_main
>     hook_generator=hook_generator,
>   File "/private/home/lucaspc/repos/vissl/vissl/engines/train.py", line 93, in train_main
>     hooks = hook_generator(cfg)
>   File "/private/home/lucaspc/repos/vissl/vissl/hooks/__init__.py", line 126, in default_hook_generator
>     wandb_hook = get_wandb_hook(cfg)
>   File "/private/home/lucaspc/repos/vissl/vissl/utils/wandb.py", line 120, in get_wandb_hook
>     name=name
>   File "/private/home/lucaspc/.conda/envs/vissl2/lib/python3.7/site-packages/wandb/sdk/wandb_init.py", line 742, in init
>     run = wi.init()
>   File "/private/home/lucaspc/.conda/envs/vissl2/lib/python3.7/site-packages/wandb/sdk/wandb_init.py", line 513, in init
>     raise UsageError(error_message)
> wandb.errors.UsageError: Error communicating with wandb process
> 
> srun: error: learnfair0928: task 0: Exited with exit code 1
> srun: launch/slurm: _step_signal: Terminating StepId=39934341.0
> ```
> 
> If anyone with distributed programming knowledge knows why this is happening, please let me know

regarding this, we need to understand what wandb requires in order to run via SLURM. Are there specific requirements or documentation on configuring wandb when using slurm?

pclucas14:
I emailed the support team at Weights and Biases, will keep you posted.

pclucas14:
After speaking with people from WandB and having them look at the log, it seems that [the following line](https://github.com/pclucas14/vissl/blob/wandb/vissl/utils/wandb.py#L115) gets called more than once, which is problematic. I need it to be called **exactly once**. Any tips on how to make that happen ? I though wrapping it in a `if is_primary()` would be sufficient, but it seems that is not the case.

QuentinDuval:
> the following line

Indeed, this is because `torch.distributed` is not initialised at this point (I got bitten by it as well). You should use `dist_rank = get_machine_local_and_dist_rank()[1]`.

prigoyal:
> > the following line
> 
> Indeed, this is because `torch.distributed` is not initialised at this point (I got bitten by it as well). You should use `dist_rank = get_machine_local_and_dist_rank()[1]`.
> 
> @prigoyal I think we need to rework this a bit, and either:
> 
> * introduce a `is_primary` function in VISSL that does the right thing (call `torch.distributed` if initialised and do as `get_machine_local_and_dist_rank` otherwise)
> * or move the creation of the hook after the initialisation of `torch.distributed`
> 
> What you do both think?

indeed. THere is already a `is_primary()` function https://github.com/facebookresearch/vissl/blob/master/vissl/hooks/tensorboard_hook.py#L7 . Does this help? :)

pclucas14:
> > > the following line
> > 
> > 
> > Indeed, this is because `torch.distributed` is not initialised at this point (I got bitten by it as well). You should use `dist_rank = get_machine_local_and_dist_rank()[1]`.
> > @prigoyal I think we need to rework this a bit, and either:
> > 
> > * introduce a `is_primary` function in VISSL that does the right thing (call `torch.distributed` if initialised and do as `get_machine_local_and_dist_rank` otherwise)
> > * or move the creation of the hook after the initialisation of `torch.distributed`
> > 
> > What you do both think?
> 
> indeed. THere is already a `is_primary()` function https://github.com/facebookresearch/vissl/blob/master/vissl/hooks/tensorboard_hook.py#L7 . Does this help? :)

I should specify that I am indeed using `is_primary` from `classy_vision.generic.distributed_util`

QuentinDuval:
> indeed. THere is already a `is_primary()` function https://github.com/facebookresearch/vissl/blob/master/vissl/hooks/tensorboard_hook.py#L7 . Does this help? :)

It will not help: it's the very function to avoid in that case because `torch.distributed` is not initialised up to that point and so it will return 0 for all callers => multiple workers will call `wandb.init`.

@pclucas14 Could you check that by replacing `is_primary` with `0 == get_machine_local_and_dist_rank()[1]` your problem of init is solved?

prigoyal:
> The function we should create would look like this:
> 
> ```
> def is_primary():
>     if not dist.is_initialized():
>         dist_rank = get_machine_local_and_dist_rank()[1]
>         return dist_rank == 0
>    else:
>        return get_rank() == 0  # get rank of classy-vision
> ```

thank you so much @pclucas14 for clarification and fully agree @QuentinDuval . I think we should go for the above function and implement it in `vissl/utils/misc.py` . We should then switch the `is_primary()` call everywhere in VISSL from classy vision to our own . @pclucas14 , if you wish, this broader change can be taken care of by us in a separate PR and in this PR, you can create a `is_primary()` as Quentin suggested above and use it where you need. :)

PR #274, opened 2021-04-06
Fix typos in bash commands in linear benchmark doc

iseessel:


prigoyal:
thank you for catching these! :) :) let's import to phabricator and land this :)

PR #275, opened 2021-04-06
[Draft] Memory profiling

QuentinDuval:
A first version of the memory profiling, tracking the memory used through the forward/backward passes, with a breakdown of the memory dedicated to activations.

- [ ] Define the test plan
- [ ] Provide example curves and data output
- [ ] Rework the commit history


QuentinDuval:
Will be moving this PR to ssl_scaling.

PR #277, opened 2021-04-07
add requirements.txt to MANIFEST, in support of making pip package by…

adamydwang:
add requirements.txt to MANIFEST, in support of making pip package by 'python setup.py sdist', so it's easy to install by 'pip install vissl-0.1.5.tar.gz'


prigoyal:
thank you @adamydwang :) this is very valuable contribution. :)

PR #283, opened 2021-04-08
[MLH] Pre-commit Hooks Configuration Changes

grace-omotoso:
- [x] Update pre-commit steps and isort configuration
- [x] Update CI test command

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -12,3 +12,4 @@ lines_after_imports = 2
 ensure_newline_before_comments = True
 no_lines_before = FIRSTPARTY
 skip_glob = build/*,third-party/*,.github/*
+known_third_party = extra_scripts
--- a/setup.cfg
+++ b/setup.cfg
@@ -9,5 +9,6 @@ line_length = 88
 multi_line_output = 3
 use_parentheses = True
 lines_after_imports = 2
-known_third_party = PIL,bs4,classy_vision,cv2,detectron2,fairscale,faiss,fvcore,hydra,mock,nbconvert,nbformat,numpy,omegaconf,parameterized,pkg_resources,pycocotools,recommonmark,run_distributed_engines,scipy,setuptools,sklearn,sphinx_rtd_theme,tabulate,torch,torchvision,tqdm,utils
+ensure_newline_before_comments = True
+no_lines_before = FIRSTPARTY
 skip_glob = build/*,third-party/*,.github/*
--- a/tools/train_svm_low_shot.py
+++ b/tools/train_svm_low_shot.py
@@ -6,11 +6,11 @@ import sys
 from argparse import Namespace
 from typing import Any, List
 
-from hydra.experimental import compose, initialize_config_module
 from extra_scripts.create_low_shot_samples import (
     generate_low_shot_samples,
     generate_places_low_shot_samples,
 )
+from hydra.experimental import compose, initialize_config_module
 from vissl.config import AttrDict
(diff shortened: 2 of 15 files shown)


tools/train_svm_low_shot.py
```diff
@@ -6,23 +6,19 @@
 from argparse import Namespace
 from typing import Any, List
 
+from hydra.experimental import compose, initialize_config_module
```
prigoyal, line 9: @akainth015 , @grace-omotoso , do we know what triggered this change? it doesn't seem aligned with the linter to me.

prigoyal (changes requested):
thank you so much. Can we take a look at one comment inline? :)

tools/train_svm_low_shot.py
akainth015: Seems correct to me, moved from line 13 to up here, probably because hydra is considered 3rd party

prigoyal: gotcha, thank you for clarifying. This currently conflicts with the linter we have at facebook which expects the alphabetical ordering. Is it possible to address this?

setup.cfg
```diff
@@ -9,5 +9,6 @@ line_length = 88
 multi_line_output = 3
 use_parentheses = True
 lines_after_imports = 2
-known_third_party = PIL,bs4,classy_vision,cv2,detectron2,fairscale,faiss,fvcore,hydra,mock,nbconvert,nbformat,numpy,omegaconf,parameterized,pkg_resources,pycocotools,recommonmark,run_distributed_engines,scipy,setuptools,sklearn,sphinx_rtd_theme,tabulate,torch,torchvision,tqdm,utils
```
prigoyal, line 12: is it possible to revert this?

setup.cfg
```diff
@@ -9,5 +9,6 @@ line_length = 88
...
+ensure_newline_before_comments = True
+no_lines_before = FIRSTPARTY
```
prigoyal, line 13: I would recommend to not introduce the new rule in this PR.

akainth015: I added these rules to minimize the effect the upgrade had on the existing files.

setup.cfg
akainth015: Yes, but since that was generated by `seed-isort-config` it seemed better to remove at, as directed by the [`isort` documentation](https://pycqa.github.io/isort/docs/configuration/pre-commit/#seed-isort-config).

tools/train_svm_low_shot.py
akainth015: Hm, does that mean alphabetical within the same block of imports? We could remove the `no_lines_before = FIRSTPARTY` directive, perhaps that would solve the issue? Although it would result in a lot of changes to existing files.

prigoyal:
thank you! I'll patch this PR today and test it and will aim to land it/merge it.

tools/train_svm_low_shot.py
iseessel: Hi @akainth015 I am new to the vissl team at FB. This PR looks great -- it will be a huge help to us!! 

I think this may have been caused by removing the `known_third_party above`?

akainth015: Hi @iseessel, nice to meet you! Glad that this will be useful

We could add `extra_scripts` to the `known_third_party` option; that would address the issue. Semantically, is that acceptable?

prigoyal (approved):

PR #284, opened 2021-04-08
[MLH Phase 2] Refactor Logging System

akainth015:
The new logging system will be based around an event storage for each `SelfSupervisionTask` and events will be logged to their corresponding output device by `VisslEventWriter`s like the `JsonWriter` and `TensorboardWriter`

- [x] https://github.com/MLH-Fellowship/vissl/issues/11
- [ ] https://github.com/MLH-Fellowship/vissl/issues/16


pranavsinghps1:
@iseessel I see some of the comments are unaddressed, do we need to work on these? Lmk if there is anything I could do!

PR #298, opened 2021-04-20
Make dataloaders configurable in anticipation for OnBoxDataloader integration

iseessel:
Summary:
In anticipation for integrating the OnBoxDataLoader, we need to make the dataloader configurable. The idea is to use a factory class, to keep the logic in #get_loader concise.

I also propose we follow the same pattern as below for the dataset configuration, as we won't be able to use GenericSSLDataset for OnBoxDataLoader.

See https://fb.quip.com/587dAkUb7pyM for more information.

Differential Revision: D27888139


prigoyal:
branch abandoned in favor of another changeset. closing this :)

PR #317, opened 2021-05-10
Introduce Hive Dataset and New Dataloader

iseessel:
Summary:
This PR introduces AutoOnBoxEverstore dataset support, which accesses everstore images through a hive table. See [Important Docs here](https://fb.quip.com/587dAkUb7pyM) for more background.

Features: Should (hopefully) support all trainings/evaluations (swav+simclr tested).

1. Multi-epoch training with seeded shuffling.
2. Custom transformations and collators.
3. Dataloader state checkpointing.
4. Support for distributed training with DDP.

Limitations
1. Multiple data sources not supported if using hive.
2. Must have label inside hive table.
3. Must specify DATA_LIMIT, will not read entire table.
4. More?

Next Steps
1. Run full 8node, 8gpu, 100epoch SwaV train/eval with imagenet to test accuracies/speed of trainings.
2. Create unit and integration tests.
3. Cleanup code: Comments, types, asserts, etc.,
4. Create followup diff for "multiple_phases_per_epoch" support (important for large datasets).
5. Post about new feature in workplace.

Differential Revision: D28080328

### DIFF
--- a/vissl/data/ssl_dataset.py
+++ b/vissl/data/ssl_dataset.py
@@ -9,11 +9,14 @@ from typing import Callable, Dict, Set
 import numpy as np
 from classy_vision.generic.distributed_util import get_world_size
 from fvcore.common.file_io import PathManager
-from torch.utils.data import Dataset
 from vissl.config import AttrDict
 from vissl.data import dataset_catalog
-from vissl.data.data_helper import balanced_sub_sampling, unbalanced_sub_sampling
+from vissl.data.data_helper import (
+    balanced_sub_sampling,
+    unbalanced_sub_sampling,
+)
 from vissl.data.ssl_transforms import get_transform
+from vissl.data.vissl_dataset_base import VisslDatasetBase
 from vissl.utils.env import get_machine_local_and_dist_rank
 
 
@@ -33,7 +36,7 @@ def _convert_lbl_to_long(lbl):
     return out_lbl
 
 
-class GenericSSLDataset(Dataset):
+class GenericSSLDataset(VisslDatasetBase):
     """
     Base Self Supervised Learning Dataset Class.
 
@@ -74,6 +77,7 @@ class GenericSSLDataset(Dataset):
         split: str,
         dataset_source_map: Dict[str, Callable],
         data_sources_with_subset: Set[str],
+        **kwargs,
     ):
         self.cfg = cfg
         self.split = split
@@ -408,3 +412,9 @@ class GenericSSLDataset(Dataset):
         The global batch size across all the trainers
         """
         return self.get_batchsize_per_replica() * get_world_size()
+
+    def post_process_batch(self, batch_data):
+        """
+        Identity function. Other base datasets may need to #post_process_batch.
+        """
+        return batch_data
--- a/vissl/data/ssl_transforms/__init__.py
+++ b/vissl/data/ssl_transforms/__init__.py
@@ -4,168 +4,10 @@
 # LICENSE file in the root directory of this source tree.
 
 from pathlib import Path
-from typing import Any, Dict
 
 import torchvision.transforms as pth_transforms
-from classy_vision.dataset.transforms import build_transform, register_transform
-from classy_vision.dataset.transforms.classy_transform import ClassyTransform
 from classy_vision.generic.registry_utils import import_all_modules
-
-
-# Below the transforms that require passing the labels as well. This is specifc
-# to SSL only where we automatically generate the labels for training. All other
-# transforms (including torchvision) require passing image only as input.
-_TRANSFORMS_WITH_LABELS = ["ImgRotatePil", "ShuffleImgPatches"]
-_TRANSFORMS_WITH_COPIES = [
-    "ImgReplicatePil",
-    "ImgPilToPatchesAndImage",
-    "ImgPilToMultiCrop",
-]
-_TRANSFORMS_WITH_GROUPING = ["ImgPilMultiCropRandomApply"]
-
-
-# we wrap around transforms so that they work with the multimodal input
-@register_transform("SSLTransformsWrapper")
-class SSLTransformsWrapper(ClassyTransform):
-    """
-    VISSL wraps around transforms so that they work with the multimodal input.
-    VISSL supports batches that come from several datasets and sources. Hence
-    the input batch (images, labels) always is a list.
-
-    To apply the user defined transforms, VISSL takes "indices" as input which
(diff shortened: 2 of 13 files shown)


vissl/data/ssl_transforms/tensor_to_img_pil.py
```diff
@@ -0,0 +1,45 @@
...
+    def __call__(self, sample):
+        # Convert tensors to image pil for subsequent transforms.
+        for i, tensor in enumerate(sample["data"]):
+            sample["data"][i] = self._tensor_to_img_pil(tensor)
+
+        sample = self._reformat_sample(sample)
+
+        return sample
+
+    def _tensor_to_img_pil(self, tensor):
```
QuentinDuval, line 26: There is a torchvision transformation that might be used here (to move from a tensor to a PIL image): https://pytorch.org/vision/stable/_modules/torchvision/transforms/transforms.html#ToPILImage

vissl/trainer/train_steps/standard_train_step.py
```diff
@@ -118,6 +118,7 @@ def standard_train_step(task):
     with PerfTimer("read_sample", perf_stats):
         sample = next(task.data_iterator)
 
+    sample = task.post_process_batch(sample)
```
QuentinDuval, line 121: 👍

vissl/trainer/train_task.py
```diff
@@ -279,43 +277,70 @@ def _build_phases(self):
...
                 device=self.device,
                 sampler_seed=sampler_seed,
+                split=split.lower(),
             )
             for split in self.available_splits
         }
 
         return loaders
 
+    def build_dataloader_iterator(self, phase_type: str):
```
QuentinDuval, line 324: 👍

vissl/trainer/train_task.py
```diff
@@ -696,7 +703,37 @@ def prepare(self, pin_memory: bool = False):
...
+            pin_memory=pin_memory, current_phase_id=current_phase_id
+        )
+        task = self._update_classy_state(vissl_state_dict)
+
+        return task
+
+    def set_epoch(self, phase_type: str, epoch: int, start_iter: int):
+        if hasattr(self.dataloaders[phase_type], "sampler"):
+            sampler = self.dataloaders[phase_type].sampler
+            # (Re-)Shuffle data: set epoch of distributed (or fairstore) sampler.
```
QuentinDuval, line 719: nit: the following lines could be factorized with below

```
if hasattr(sampler, "set_epoch"):
    sampler.set_epoch(epoch)
# Resume from the iteration if valid
if hasattr(sampler, "set_start_iter"):
    sampler.set_start_iter(start_iter)
```

Another idea would be to introduce a "tagging interface": instead of checking the presence of method, do a `isinstance` check of an interface, that contains the functions you want. The main advantage is that the name of the interface might bring some more additional information.

vissl/data/ssl_dataset.py
```diff
@@ -33,7 +36,7 @@ def _convert_lbl_to_long(lbl):
     return out_lbl
 
 
-class GenericSSLDataset(Dataset):
+class GenericSSLDataset(VisslDatasetBase):
```
QuentinDuval, line 39: 👍 👍

vissl/data/ssl_transforms/__init__.py
```diff
@@ -4,185 +4,9 @@
...
-        transform = SSLTransformsWrapper.from_config(transform_config)
-        output_transforms.append(transform)
-    return pth_transforms.Compose(output_transforms)
-
-
 FILE_ROOT = Path(__file__).parent
-import_all_modules(FILE_ROOT, "vissl.data.ssl_transforms")
 
-__all__ = ["SSLTransformsWrapper", "get_transform"]
+import_all_modules(FILE_ROOT, "vissl.data.ssl_transforms")
```
QuentinDuval, line 12: 👍 👍

vissl/data/vissl_dataset_base.py
```diff
@@ -0,0 +1,44 @@
+from typing import Callable, Dict, Set
+
+import torch
+from torch.utils.data import Dataset
+from vissl.config import AttrDict
+
+
+class VisslDatasetBase(Dataset):
```
QuentinDuval, line 8: nit: consider using `abc` to define the abstract methods

```
class Base(abc.ABC):
   @abc.abstractmethod
   def f(self):
      ...
```

This makes the class and sub class not instantiable if they do not define the missing functions

vissl/data/ssl_transforms/tensor_to_img_pil.py
```diff
@@ -0,0 +1,45 @@
...
+
+from io import BytesIO
+
+from classy_vision.dataset.transforms import register_transform
+from classy_vision.dataset.transforms.classy_transform import ClassyTransform
+from PIL import Image
+
+
+@register_transform("TensorToImgPil")
+class TensorToImgPil(ClassyTransform):
```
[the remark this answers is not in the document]
iseessel: Yeah -- not really. It should probably be moved to internal.

I refrained from doing this because it would then require some refactoring of the code here: https://github.com/facebookresearch/vissl/pull/317/files#diff-82aa56b0a134fa104a6d729561ff3aaf76e2b69c735037d7c8858480dafc5622R140 to be because this class would no longer be importable.

vissl/data/vissl_dataset_base.py
iseessel: Yeah, I wasn't sure whether or not we needed the datasets to inherit from the torch Dataset, I'll test.

vissl/trainer/train_task.py
iseessel: Yep indeed, good catch (this code is existing, just moved).

IMO, I am not loving this `has_attr` pattern in general. I would prefer an object oriented interface w/ default no-op methods that can be overwritten.

vissl/data/ssl_transforms/tensor_to_img_pil.py
iseessel: So I tried this, it doesn't work unfortunately. The tensor here is flattened, and torchvision expects a C x W x H tensor.

vissl/data/ssl_transforms/tensor_to_img_pil.py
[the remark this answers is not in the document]
iseessel: I move to internal in subsequent PR.

PR #325, opened 2021-05-25
Continuous evaluations init commit

iseessel:
Create a script that continuously evaluates benchmarks as they become available from a pretraining.

Screen Shot 2021-06-02 at 10 22 37 AM
Screen Shot 2021-06-02 at 10 22 59 AM

Next Steps:
1. Deal with sharded checkpoints and their conversion
1. Improve max_iteration logic
1. Extend to FB infra.
1. Write unit tests
1. Think about how these tricky evaluation tests: https://github.com/facebookresearch/vissl/pull/325#issuecomment-853047525
1. Try not to replicate so much logic in the class (e.g. get path names from vissl code, requires some refactoring).
1. Look into email notifications.

Testing:

1. Run 8node Swav with 10 epochs with 3 different benchmark evaluations with different resource requirements. SUCCESS. 

json config: 

```
{
    "params": {
           "training_checkpoint_dir": "/checkpoint/iseessel/vissl/2021-06-09-11-19-12/checkpoints",
           "benchmarks": [
               {
                   "evaluation_name": "clevr_count_linear",
                   "config_files": [
                       "config=config_local/eval_resnet_8gpu_transfer_clevr_count_linear_benchmark_suite_scheduler_test.yaml"
                   ]
               },
               {
                   "evaluation_name": "clevr_dist_linear",
                   "config_files": [
                       "config=config_local/eval_resnet_8gpu_transfer_clevr_dist_linear_benchmark_suite_scheduler_test.yaml"
                   ]
               },
               {
                   "evaluation_name": "in1k_linear",
                   "config_files": [
                       "config=config_local/eval_resnet_8gpu_transfer_in1k_linear_benchmark_suite_scheduler_test.yaml"
                   ]
               }
           ],
           "evaluation_iter_freq": 600,
           "evaluation_phase_freq": 2,
           "evaluate_final_phase": true,
           "autoload_slurm_evaluator_checkpoint": false,
           "slurm_evaluator_checkpoint": null,
           "auto_retry_evaluations": true,
           "retry_evaluation_job_ids": [],
           "max_retries": 3,
           "pytorch_ports": [40050, 40051, 40052, 40053, 40054, 40055, 40056, 40057, 40058, 40059, 40060, 40061, 40062, 40063]
       },
       "slurm_options": {
           "PARTITION": "learnfair"
       }
}
```

Example snippet from `evaluation_metrics.json`:

```
{
    "model_final_checkpoint_phase9": [
        {
            "checkpoint_dir": "/checkpoint/iseessel/vissl/2021-06-09-11-19-12/checkpoints/evaluations/model_final_checkpoint_phase9/clevr_count_linear/checkpoints",
            "config_files": [
                "config=config_local/eval_resnet_8gpu_transfer_clevr_count_linear_benchmark_suite_scheduler_test.yaml",
                "hydra.run.dir='/checkpoint/iseessel/vissl/2021-06-09-11-19-12/checkpoints/evaluations/model_final_checkpoint_phase9/clevr_count_linear'",
                "config.CHECKPOINT.DIR='/checkpoint/iseessel/vissl/2021-06-09-11-19-12/checkpoints/evaluations/model_final_checkpoint_phase9/clevr_count_linear/checkpoints'",
                "config.SLURM.LOG_FOLDER='/checkpoint/iseessel/vissl/2021-06-09-11-19-12/checkpoints/evaluations/model_final_checkpoint_phase9/clevr_count_linear'",
                "config.SLURM.LOG_FOLDER='/checkpoint/iseessel/vissl/2021-06-09-11-19-12/checkpoints/evaluations/model_final_checkpoint_phase9/clevr_count_linear'",
                "config.SLURM.USE_SLURM=true",
                "config.MODEL.WEIGHTS_INIT.PARAMS_FILE='/checkpoint/iseessel/vissl/2021-06-09-11-19-12/checkpoints/model_final_checkpoint_phase9.torch'"
            ],
            "evaluation_name": "clevr_count_linear",
            "job_id": "42410489",
            "metrics": {
                "test_accuracy_list_meter_top_1_res5": {
                    "iteration": 822,
                    "metric": 34.62,
                    "train_phase_idx": 2
                },
                "train_accuracy_list_meter_top_1_res5": {
                    "iteration": 822,
                    "metric": 33.8514,
                    "train_phase_idx": 2
                }
            },
            "num_retries": 1,
            "slurm_checkpoint_dir": "/checkpoint/iseessel/vissl/2021-06-09-11-19-12/checkpoints/evaluations/model_final_checkpoint_phase9/clevr_count_linear/checkpoints",
            "slurm_log_dir": "/checkpoint/iseessel/vissl/2021-06-09-11-19-12/checkpoints/evaluations/model_final_checkpoint_phase9/clevr_count_linear",
            "slurm_state": "COMPLETED",
            "weights_init_params_file": "/checkpoint/iseessel/vissl/2021-06-09-11-19-12/checkpoints/model_final_checkpoint_phase9.torch"
        }, ...
```

The following hold: 
1. Training completes appropriately, w/o errors.
1. Able to resume checkpoints.
1. Evaluation folder structure is as expected above.
1. Best Metrics are extracted.

### DIFF
--- a/dev/launch_benchmark_suite_scheduler_slurm.sh
+++ b/dev/launch_benchmark_suite_scheduler_slurm.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+# Copyright (c) Facebook, Inc. and its affiliates.
+
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+# This benchmark suite script launches a benchmark suite scheduler slurm job.
+# The job takes an absolute json config path (see benchmark_suite_scheduler_template.json for info)
+# The job continuously monitors training benchmarks, and dynamically launches evaluation jobs
+# and amalgamates the results.
+
+######################### EXAMPLE USAGE #################################
+
+# cd into vissl root directory.
+#
+# bash ./dev/launch_benchmark_suite_scheduler_slurm.sh /path/to/benchmark_suite_scheduler.json
+
+# See benchmark_suite_scheduler_template.json or for config information or slurm_evaluator.py for class structure.
+######################### INPUT PARAMS ##################################
+
+FILE=( "$@" )
+
+####################### setup experiment dir ###################################
+
+# create a temporary experiment folder to run the SLURM job in isolation
+RUN_ID=$(date +'%Y-%m-%d-%H-%M-%S')
+EXP_ROOT_DIR="/checkpoint/$USER/vissl/$RUN_ID"
+
+echo "EXP_ROOT_DIR: $EXP_ROOT_DIR"
+echo "CONFIG_FILE: ${FILE[0]}"
+
+rm -rf "$EXP_ROOT_DIR"
+mkdir -p "$EXP_ROOT_DIR"
+cp -r . "$EXP_ROOT_DIR"
+
+####################### setup experiment dir ###################################
+export PYTHONPATH="$EXP_ROOT_DIR/:$PYTHONPATH"
+python -u "$EXP_ROOT_DIR/tools/launch_benchmark_suite_scheduler_slurm.py" \
+    "${FILE[@]}"
--- a/tools/launch_benchmark_suite_scheduler_slurm.py
+++ b/tools/launch_benchmark_suite_scheduler_slurm.py
@@ -1,0 +1,107 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import sys
+
+import pkg_resources
+import submitit
+from fvcore.common.file_io import PathManager
+from vissl.config.attr_dict import AttrDict
+from vissl.utils.benchmark_suite_scheduler import BenchmarkSuiteScheduler
+from vissl.utils.hydra_config import is_hydra_available
+from vissl.utils.io import load_file
+from vissl.utils.misc import recursive_dict_merge
+from vissl.utils.slurm import is_submitit_available
+
+
+# Default config options
+default_config_file = pkg_resources.resource_filename(
+    "dev", "benchmark_suite/benchmark_suite_scheduler_defaults.json"
+)
+_DEFAULT_CONFIG = load_file(default_config_file)
+
+
+class SlurmEvaluatorJob:
+    """
+    The slurm evaluator job is a thin wrapper around BenchmarkSuiteScheduler
+    used by submitit. It's main function is to run multiple evaluations
+    on a single training.
+    """
+
+    def __init__(self, benchmark_suite_scheduler: BenchmarkSuiteScheduler):
+        self.benchmark_suite_scheduler = benchmark_suite_scheduler
+
+    def __call__(self):
+        self.benchmark_suite_scheduler.evaluate()
+
+    def checkpoint(self):
+        """
+        This method is called whenever a job is pre-empted, timedout, etc,.
+        Here we save the evaluation benchmarks, so that we can reload them
+        and continue where we left off.
+        """
+        self.benchmark_suite_scheduler.save_evaluation_benchmarks()
+        # Forces the benchmark_suite_scheduler to automatically reload it's
+        # checkpoint, the benchmark results.
+        self.benchmark_suite_scheduler.autoload_benchmark_suite_scheduler_checkpoint = (
+            True
+        )
+
+        trainer = SlurmEvaluatorJob(
+            benchmark_suite_scheduler=self.benchmark_suite_scheduler
+        )
+        return submitit.helpers.DelayedSubmission(trainer)
+
+
+def launch_benchmark_suite_scheduler(config_file):
+    assert PathManager.exists(config_file), "Slurm evaluator config file must exist"
+
+    user_config = load_file(config_file)
+    config = _DEFAULT_CONFIG.copy()
+    recursive_dict_merge(config, user_config)
+
+    benchmark_suite_scheduler = BenchmarkSuiteScheduler(**config["params"])
+    benchmark_suite_scheduler_job = SlurmEvaluatorJob(
+        benchmark_suite_scheduler=benchmark_suite_scheduler
+    )
+    executor = submitit.AutoExecutor(folder=benchmark_suite_scheduler.evaluation_dir())
+
+    assert "slurm_options" in config, "slurm_options must be specified"
+    assert (
+        "PARTITION" in config["slurm_options"]
+    ), "slurm_options.PARTITION is a required field to launch the benchmark suite on slurm"
+
+    slurm_options = AttrDict(config["slurm_options"])
+    executor.update_parameters(
+        name=slurm_options.NAME,
+        slurm_comment=slurm_options.COMMENT,
+        slurm_partition=slurm_options.PARTITION,
+        slurm_constraint=slurm_options.CONSTRAINT,
+        timeout_min=slurm_options.TIMEOUT_MIN,
+        nodes=1,
+        cpus_per_task=slurm_options.CPUS_PER_TASK,
+        tasks_per_node=1,
+        mem_gb=slurm_options.MEM_GB,
+        slurm_additional_parameters=slurm_options.ADDITIONAL_PARAMETERS,
+    )
+
+    job = executor.submit(benchmark_suite_scheduler_job)
+    print(f"SUBMITTED EVALUATION JOB: {job.job_id}")
+
+
+if __name__ == "__main__":
+    """
+    Example usage:
+    python -u "./vissl/engines/benchmark_suite_scheduler.py" \
+        "/path/to/benchmark_suite_scheduler_example.json"
+    """
+    assert is_hydra_available(), "Make sure to install hydra"
+
+    assert (
+        is_submitit_available()
+    ), "Please 'pip install submitit' to schedule jobs on SLURM"
+
+    config_file = sys.argv[1]
+    launch_benchmark_suite_scheduler(config_file)
--- a/vissl/utils/misc.py
+++ b/vissl/utils/misc.py
@@ -3,10 +3,13 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
+import collections
 import logging
 import os
 import random
 import tempfile
+import time
+from functools import partial, wraps
 
 import numpy as np
 import pkg_resources
@@ -308,3 +311,113 @@ class set_torch_seed(object):
 
     def __exit__(self, *exc):
         set_rng_state(self.rng_state)
+
+
+# Credit: https://stackoverflow.com/questions/42521549/retry-function-in-python
+def retry(func=None, exception=Exception, n_tries=5, delay=5, backoff=1, logger=False):
+    """Retry decorator with exponential backoff.
+
+    Parameters
+    ----------
+    func : typing.Callable, optional
+        Callable on which the decorator is applied, by default None
+    exception : Exception or tuple of Exceptions, optional
+        Exception(s) that invoke retry, by default Exception
+    n_tries : int, optional
+        Number of tries before giving up, by default 5
+    delay : int, optional
+        Initial delay between retries in seconds, by default 5
+    backoff : int, optional
+        Backoff multiplier e.g. value of 2 will double the delay, by default 1
+    logger : bool, optional
+        Option to log or print, by default False
+
+    Returns
+    -------
+    typing.Callable
+        Decorated callable that calls itself when exception(s) occur.
+
+    Examples
+    --------
+    >>> import random
+    >>> @retry(exception=Exception, n_tries=4)
+    ... def test_random(text):
+    ...    x = random.random()
+    ...    if x < 0.5:
+    ...        raise Exception("Fail")
+    ...    else:
+    ...        print("Success: ", text)
+    >>> test_random("It works!")
+    """
+
+    if func is None:
+        return partial(
+            retry,
+            exception=exception,
+            n_tries=n_tries,
+            delay=delay,
+            backoff=backoff,
+            logger=logger,
+        )
+
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        ntries, ndelay = n_tries, delay
+
+        while ntries > 1:
+            try:
+                return func(*args, **kwargs)
+            except exception as e:
+                msg = f"{str(e)}, Retrying in {ndelay} seconds..."
+                if logger:
+                    logging.warning(msg)
+                else:
+                    print(msg)
+                time.sleep(ndelay)
+                ntries -= 1
+                ndelay *= backoff
+
+        return func(*args, **kwargs)
+
+    return wrapper
+
+
+# Credit: https://stackoverflow.com/questions/6027558/flatten-nested-dictionaries-compressing-keys # NOQA
+def flatten_dict(d: dict, parent_key="", sep="_"):
+    """
+    Flattens a dict, delimited with a '_'. For example the input:
+    {
+        'top_1': {
+            'res_5': 100
+        }
+    }
+
+    will return:
+
+    {
+        'top_1_res_5': 100
+    }
+    """
+    items = []
+    for k, v in d.items():
+        new_key = parent_key + sep + k if parent_key else k
+        if isinstance(v, collections.MutableMapping):
+            items.extend(flatten_dict(v, new_key, sep=sep).items())
+        else:
+            items.append((new_key, v))
+    return dict(items)
+
+
+# Credit: https://stackoverflow.com/questions/7204805/how-to-merge-dictionaries-of-dictionaries
+def recursive_dict_merge(dict1, dict2):
+    """
+    Recursively merges dict2 into dict1
+    """
+    if not isinstance(dict1, dict) or not isinstance(dict2, dict):
+        return dict2
+    for k in dict2:
+        if k in dict1:
+            dict1[k] = recursive_dict_merge(dict1[k], dict2[k])
+        else:
+            dict1[k] = dict2[k]
+    return dict1
--- a/.gitignore
+++ b/.gitignore
@@ -97,5 +97,5 @@ website/pages/tutorials/*
 **/.ipynb_checkpoints/**
 
 # Configs for local development
-configs/config_local/*
+configs/config/config_local/*
 train_config.yaml
--- a/dev/benchmark_suite/benchmark_suite_scheduler_defaults.json
+++ b/dev/benchmark_suite/benchmark_suite_scheduler_defaults.json
@@ -1,0 +1,22 @@
+{
+    "params": {
+           "evaluation_iter_freq": -1,
+           "evaluation_phase_freq": -1,
+           "evaluate_final_phase": true,
+           "autoload_slurm_evaluator_checkpoint": false,
+           "slurm_evaluator_checkpoint": null,
+           "auto_retry_evaluations": false,
+           "retry_evaluation_job_ids": [],
+           "max_retries": 3,
+           "pytorch_ports": [40050]
+       },
+       "slurm_options": {
+            "NAME": "vissl",
+            "COMMENT": "vissl evaluation job",
+            "CONSTRAINT": "",
+            "TIMEOUT_MIN": 4320,
+            "CPUS_PER_TASK": 8,
+            "MEM_GB": 16,
+            "ADDITIONAL_PARAMETERS": {}
+        }
+}
--- a/dev/benchmark_suite/benchmark_suite_scheduler_template.json
+++ b/dev/benchmark_suite/benchmark_suite_scheduler_template.json
@@ -1,0 +1,33 @@
+{
+    "params": {
+           "training_checkpoint_dir": "(str) Training checkpoint directory. That is the CHECKPOINT.DIR of the training config",
+           "benchmarks": [
+               {
+                   "evaluation_name": "(str) Name of benchmark for convenience",
+                   "config_files": [
+                       "config=path/to/evaluation/config",
+                       "config.OVERRIDES=new_value"
+                   ]
+               }
+           ],
+           "evaluation_iter_freq": "(int, default=-1) Evaluate the checkpoint every N iterations",
+           "evaluation_phase_freq": "(int, default=-1) Evaluate the checkpoint every N phases",
+           "evaluate_final_phase": "(bool, default=True) Evaluate the final phase",
+           "autoload_slurm_evaluator_checkpoint": "(bool, default=False) Whether or not to automatically load the benchmark checkpoint",
+           "slurm_evaluator_checkpoint": "(str, default=None) Path to load the benchmark checkpoint",
+           "auto_retry_evaluations": "(bool, default=False) Whether or not to automatically retry the evaluations",
+           "retry_evaluation_job_ids": "(array[int], default=[]) Array of job_ids to retry",
+           "max_retries": "(int, default=3) Maximum number of retries",
+           "pytorch_ports": "(List[int], default=[40500]) List of pytorch ports to cycle through as you are launching your evaluations, in order to prevent Pytorch DDP port colissions."
+       },
+       "slurm_options": {
+            "PARTITION": "(str) Partition",
+            "NAME": "(str, default=vissl) Name of slurm job",
+            "COMMENT": "(str, default=vissl evaluation job) Comment of slurm job",
+            "CONSTRAINT": "(str, default='') Constraing of slurm job",
+            "TIMEOUT_MIN": "(int, default=72 * 60) Minimum amount of minutes to timeout",
+            "CPUS_PER_TASK": "(int, default=8) Numer of cpus per task.",
+            "MEM_GB": "(int, default=32) Amount of RAM to request from slurm",
+            "ADDITIONAL_PARAMETERS": "(Dict[[str, Any]], default={}) Any default slurm options to pass to submitit"
+    }
+}
--- a/vissl/hooks/log_hooks.py
+++ b/vissl/hooks/log_hooks.py
@@ -245,6 +245,11 @@ class LogLossLrEtaHook(ClassyHook):
                     "eta": eta_string,
                     "peak_mem(M)": peak_mem_used,
                 }
+
+                if iteration == 1:
+                    # Set max iterations. Currently used in benchmark_suite_scheduler.py
+                    log_data["max_iterations"] = task.max_iteration
+
                 if self.btime_freq and len(batch_times) >= self.btime_freq:
                     rolling_avg_time = (
                         sum(batch_times[-self.btime_freq :]) / self.btime_freq
--- a/vissl/utils/benchmark_suite_scheduler.py
+++ b/vissl/utils/benchmark_suite_scheduler.py
@@ -1,0 +1,618 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+import collections
+import json
+import logging
+import os
+import time
+from datetime import datetime
+from pathlib import Path
+from typing import List
+
+import submitit
+from fvcore.common.file_io import PathManager
+from hydra.experimental import compose, initialize_config_module
+from vissl.config.attr_dict import AttrDict
+from vissl.utils.distributed_launcher import launch_distributed_on_slurm
+from vissl.utils.hydra_config import convert_to_attrdict
+from vissl.utils.io import load_file, makedir
+from vissl.utils.misc import flatten_dict, retry
+
+
+"""
+This class is designed to be used to run multiple evaluations on a single (pre)training.
+Using the #evaluate method we continuously monitor training checkpoints, launch evaluations
+dynamically as they become available, and amalgamate the evaluation results as they become
+available.
+
+For SLURM usage, you should create a JSON configuration file
+(see benchmark_suite_scheduler_template.json) and use
+launch_benchmark_suite_scheduler_slurm.sh for convenience.
+"""
+
+_DEFAULT_PYTORCH_PORTS = [40050]
+# How many times to retry a slurm job submission.
+_NUM_SLURM_RETRIES = 5
+# How many seconds to sleep between iterations of the main loop.
+_SLEEP_TIME_SECONDS = 15
+# Slurm states marked as terminal. SlurmEvulator#evaluate will finish
+# once all jobs are in a terminal state.
+_SLURM_JOB_TERMINAL_STATES = [
+    "BOOT_FAIL",
+    "CANCELLED",
+    "COMPLETED",
+    "DEADLINE",
+    "FAILED",
+    "NODE_FAIL",
+    "OUT_OF_MEMORY",
+    "REVOKED",
+    "SPECIAL_EXIT",
+    "STOPPED",
+    "SUSPENDED",
+    "TIMEOUT",
+]
+# Wait for the training checkpoint folder to be available for 1 hour.
+_TRAINING_CONFIG_WAIT_SECONDS = 60 * 60
+
+
+class BenchmarkSuiteScheduler:
+    """
+    The Slurm Evaluator is a class designed to continuously monitor VISSL pretrainings
+    and launch evaluations as checkpoints become available. The method takes a
+    config dictionary consisting of the training checkpoint directory, an array of
+    benchmarks, and information on how often to evaluate the trainings.
+    """
+
+    def __init__(
+        self,
+        training_checkpoint_dir: str,
+        benchmarks: List,
+        evaluate_final_phase=True,
+        evaluation_phase_freq: int = -1,
+        evaluation_iter_freq: int = -1,
+        autoload_slurm_evaluator_checkpoint=False,
+        slurm_evaluator_checkpoint: str = None,
+        retry_evaluation_job_ids: List[int] = None,
+        auto_retry_evaluations=False,
+        max_retries=3,
+        pytorch_ports=None,
+    ):
+        """
+        Args:
+
+        training_checkpoint_dir: (str). Checkpoint directory of the training.
+                                This should match the trainings CHECKPOINT.dir
+        benchmarks: (list[dict]) Benchmarks with the following structure:
+            "config_files": [
+                    {
+                        # Path to config file.
+                        "config=test/integration_test/quick_eval_in1k_linear.yaml"
+                        # Config overrides.
+                        "config.TRAIN.DATA_LIMIT=1000",
+                        ...
+                    },
+                    ...
+                ]
+        evaluate_final_phase: (bool, optional, default=True). Whether or not to evaluate the
+                                final phase of the training.
+        evaluation_phase_freq: (int, optional, default=-1) How often to evaluate phases.
+                                Training checkpoint phase freq must evenly
+                                divide evaluation_phase_freq.
+        evaluation_iter_freq: (int, optional, default=-1) How often to evaluate iterations.
+                               Training checkpoint iteration freq must evenly
+                               divide evaluation_iter_freq.
+        autoload_slurm_evaluator_checkpoint: (bool, optional, default=False) Whether or not to
+                                              autoload slurm_evaluator Checkpoint.
+                                              This is useful when slurm evaluator job
+                                              is preempted for example.
+        slurm_evaluator_checkpoint: (str, optional, default=None) String of
+                                    slurm_evaluator checkpoint directory.
+        retry_evaluation_job_ids: (List[int], optional, default=[]) List of job_ids to retry.
+        auto_retry_evaluations: (bool, optional, default=False) Whether or not to automatically
+                                retry all failed jobs.
+        max_retries: (int, optional, default=3). Maximum number of retries.
+        pytorch_ports: (list[int], optional, default=[40050]). Ports to cycle through as
+                        you are launching your trainings.
+        """
+        self.evaluation_jobs_finished = set()
+
+        # Required Arguments
+        self.training_checkpoint_dir = training_checkpoint_dir
+        self.training_checkpoint_file = os.path.join(
+            self.training_checkpoint_dir, "train_config.yaml"
+        )
+        self.benchmarks = benchmarks
+
+        # Optional Arguments
+        self.evaluate_final_phase = evaluate_final_phase
+        self.evaluation_phase_freq = evaluation_phase_freq
+        self.evaluation_iter_freq = evaluation_iter_freq
+        self.autoload_slurm_evaluator_checkpoint = autoload_slurm_evaluator_checkpoint
+        self.slurm_evaluator_checkpoint = slurm_evaluator_checkpoint
+        self.retry_evaluation_job_ids = retry_evaluation_job_ids or []
+        self.auto_retry_evaluations = auto_retry_evaluations
+        self.max_retries = max_retries
+        self.pytorch_ports = pytorch_ports or _DEFAULT_PYTORCH_PORTS
+        self.pytorch_ports_iterable = iter(self.pytorch_ports)
+
+        self.validate()
+
+        # Will be set in #evaluate, once training_checkpoint_dir becomes available.
+        self.training_config = None
+        self.evaluation_results = None
+
+    def evaluate(self):
+        """
+        Evaluate the checkpoints. At a high level, this is the structure.
+
+        1. Load training YAML config file.
+        2. Monitor training checkpoints. When checkpoint is ready, launch the evaluation jobs.
+        3. Monitor the evaluation jobs. When evaluation jobs are complete, mark the results.
+        3. Once all jobs have been recorded, complete.
+        """
+        start_time = time.time()
+
+        # Wait for the training config to be available. This indicates the training has begun.
+        while True:
+            if time.time() - start_time > _TRAINING_CONFIG_WAIT_SECONDS:
+                raise RuntimeError(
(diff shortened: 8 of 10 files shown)


vissl/engines/slurm_evaluator.py
```diff
@@ -0,0 +1,506 @@
...
+        benchmark_result["metrics"] = None
+        benchmark_result["slurm_state"] = None
+        benchmark_result["config_files"] = benchmark["config_files"].copy()
+
+        # Hydra config information
+        weights_init_path = os.path.join(
+            self.training_config.CHECKPOINT.DIR, f"{ training_checkpoint }.torch"
+        )
+        config_files = benchmark_result["config_files"]
+        config_files.insert(
```
QuentinDuval, line 424: I think you could `append` instead of `insert(1, )` unless you are worried that these options are already in the `"config_files"` configuration?

vissl/engines/slurm_evaluator.py
```diff
@@ -0,0 +1,506 @@
...
+            1, f"config.MODEL.WEIGHTS_INIT.PARAMS_FILE='{weights_init_path}'"
+        )
+        config_files.insert(1, "config.SLURM.USE_SLURM=true")
+        config_files.insert(1, f"config.SLURM.LOG_FOLDER='{log_dir}'")
+        config_files.insert(1, f"config.CHECKPOINT.DIR='{checkpoint_dir}'")
+        config_files.insert(1, f"hydra.run.dir='{ log_dir }'")
+        config_files.insert(1, "hydra.verbose=true")
```
QuentinDuval, line 431: You could skip that one

vissl/engines/slurm_evaluator.py
```diff
@@ -0,0 +1,506 @@
...
+
+class SlurmEvaluatorJob:
+    def __init__(self, slurm_evaluator: SlurmEvaluator):
+        self.slurm_evaluator = slurm_evaluator
+
+    def __call__(self):
+        self.slurm_evaluator.evaluate()
+
+    def checkpoint(self):
+        self.slurm_evaluator.save_evaluation_results()
```
QuentinDuval, line 468: 👍

vissl/engines/slurm_evaluator.py
```diff
@@ -0,0 +1,506 @@
...
+        mem_gb=slurm_options.MEM_GB,
+        slurm_additional_parameters=slurm_options.ADDITIONAL_PARAMETERS,
+    )
+
+    job = executor.submit(slurm_evaluator_job)
+    print(f"SUBMITTED EVALUATION JOB: {job.job_id}")
+
+    return job
+
+if __name__ == "__main__":
```
QuentinDuval, line 504: The idea would be to create a new script under `extra_script` or `tool` folder (not sure which is the best) and move the `__main__` in this part.

vissl/engines/slurm_evaluator.py
iseessel: Correct, I want those already in config_files to overwrite these options. E.g. specifying specific log directory.

vissl/engines/slurm_evaluator.py
```diff
@@ -0,0 +1,506 @@
...
+                state_prev: None, state_current: { job.state }
+            """
+
+            print(log)
+
+            self.save_evaluation_results()
+
+    @retry(n_tries=_NUM_SLURM_RETRIES)
+    def _launch_slurm_job(self, args, config):
+        return launch_distributed_on_slurm(engine_name=args.engine_name, cfg=config)
```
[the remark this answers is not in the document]
iseessel: I looked into this, but didn't think it was worth the cost -- this complicates things: "jobs must have the same initial options (e.g. size, time limit, etc.), however it is possible to change some of these options after the job has begun execution using the scontrol command specifying the JobID of the array or individual ArrayJobID." (https://slurm.schedmd.com/job_array.html).

vissl/engines/slurm_evaluator.py
```diff
@@ -0,0 +1,506 @@
...
+            num_iterations = self.max_training_iterations
+        else:
+            # TODO: This can potentially be improved.
+            train_cfg = cfg.DATA.TRAIN
+            global_batch_size = (
+                train_cfg.BATCHSIZE_PER_REPLICA
+                * cfg.DISTRIBUTED.NUM_NODES
+                * cfg.DISTRIBUTED.NUM_PROC_PER_NODE
+            )
+            iterations_per_epoch = train_cfg.DATA_LIMIT // global_batch_size
```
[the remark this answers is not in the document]
iseessel: I agree this is not the best way. Let me think more about how to improve this and I can do a followup PR to improve.

tools/launch_benchmark_suite_scheduler_slurm.py
```diff
@@ -0,0 +1,83 @@
...
+from vissl.utils.io import load_file
+
+
+# Default slurm options to pass to the executor.
+_DEFAULT_SLURM_OPTIONS = {
+    "NAME": "vissl",
+    "COMMENT": "vissl evaluation job",
+    "PARTITION": "learnfair",
+    "CONSTRAINT": "",
+    "TIMEOUT_MIN": 4320,  # Timeout in minutes.
```
QuentinDuval, line 16:
```suggestion
    "TIMEOUT_MIN": 72 * 60,  # Timeout in 72 hours
```

tools/launch_benchmark_suite_scheduler_slurm.py
```diff
@@ -0,0 +1,83 @@
...
+from vissl.config.attr_dict import AttrDict
+from vissl.engines.benchmark_suite_scheduler import BenchmarkSuiteScheduler
+from vissl.utils.io import load_file
+
+
+# Default slurm options to pass to the executor.
+_DEFAULT_SLURM_OPTIONS = {
+    "NAME": "vissl",
+    "COMMENT": "vissl evaluation job",
+    "PARTITION": "learnfair",
```
QuentinDuval, line 14: Suggestion: "learnlab" is faster

QuentinDuval:
I had some suggestions for tests for the benchmark suite evaluator, which I think will allow us to exercise the most tricky cases:

- [ ] `configs/config/benchmark/linear_image_classification/voc07` because it is using SVM evaluation
- [ ] `configs/config/benchmark/nearest_neighbor/eval_resnet_8gpu_in1k_kNN.yaml` because it is using KNN evaluation
- [ ] `configs/config/benchmark/object_detection` because the evaluation is not implemented in VISSL

These uses cases do not have to be supported in this PR, but we need to think about them:
- Either to make the benchmark suite evaluator support them
- Or to refactor these benchmarks so that they fit more naturally (my preferred option)

vissl/trainer/train_task.py
```diff
@@ -322,6 +322,9 @@ def build_dataloaders(
 
         return loaders
 
+    def num_training_iterations(self):
+        return len(self.dataloaders["train"]) * self.config["OPTIMIZER"]["num_epochs"]
```
QuentinDuval, line 326: Excellent :) I love that !

dev/launch_benchmark_suite_scheduler_slurm.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+# Copyright (c) Facebook, Inc. and its affiliates.
+
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
```
prigoyal, line 6: nit: add a description here of what the purpose of this bash script is

tools/launch_benchmark_suite_scheduler_slurm.py
```diff
@@ -0,0 +1,82 @@
+import sys
```
prigoyal, line 1: nit: add the license header above

prigoyal, line 1: nit: how about renaming it to: "benchmark_suite_scheduler.py" ?

tools/launch_benchmark_suite_scheduler_slurm.py
```diff
@@ -0,0 +1,82 @@
...
+    "PARTITION": "learnlab",
+    "CONSTRAINT": "",
+    "TIMEOUT_MIN": 72 * 60,  # Timeout in minutes.
+    "CPUS_PER_TASK": 8,
+    "MEM_GB": 16,
+    "ADDITIONAL_PARAMETERS": {},
+}
+
+
+class SlurmEvaluatorJob:
```
prigoyal, line 23: nit: add a small docstring for the class. important for documentation purposes in vissl sphinx docs

tools/launch_benchmark_suite_scheduler_slurm.py
```diff
@@ -0,0 +1,82 @@
...
+from vissl.config.attr_dict import AttrDict
+from vissl.engines.benchmark_suite_scheduler import BenchmarkSuiteScheduler
+from vissl.utils.io import load_file
+
+
+# Default slurm options to pass to the executor.
+_DEFAULT_SLURM_OPTIONS = {
+    "NAME": "vissl",
+    "COMMENT": "vissl evaluation job",
+    "PARTITION": "learnlab",
```
prigoyal, line 14: nit: this is specific to FAIR  cluster. we should remove this from defaults. :)

tools/launch_benchmark_suite_scheduler_slurm.py
```diff
@@ -0,0 +1,82 @@
...
+    "CONSTRAINT": "",
```
prigoyal, line 15: nit: remove this from default. I think we should let user explicitly decide the constraint and not make default assumption

tools/launch_benchmark_suite_scheduler_slurm.py
```diff
@@ -0,0 +1,82 @@
...
+    "TIMEOUT_MIN": 72 * 60,  # Timeout in minutes.
+    "CPUS_PER_TASK": 8,
+    "MEM_GB": 16,
```
prigoyal, line 18: is it too less ?

tools/launch_benchmark_suite_scheduler_slurm.py
```diff
@@ -0,0 +1,82 @@
...
+    "ADDITIONAL_PARAMETERS": {},
```
prigoyal, line 19: nit: remove this.

vissl/trainer/train_task.py
prigoyal: nit: add an assert to check that the self.dataloaders is created and defined. Also , sometimes, users can run "TEST_ONLY" mode in VISSL so the "train" will not exist. We should handle the assumptions via asserts.

tools/launch_benchmark_suite_scheduler_slurm.py
```diff
@@ -0,0 +1,82 @@
...
+    return job
+
+
+if __name__ == "__main__":
+    """
+    Example usage:
+    python -u "./vissl/engines/benchmark_suite_scheduler.py" \
+        "/path/to/checkpoint_file/checkpoint.torch"
+    """
+    config_file = sys.argv[1]
```
prigoyal, line 81: nit: it's slightly confusing from the Example usage: the config file should be a .json or .torch or both are required?

tools/launch_benchmark_suite_scheduler_slurm.py
```diff
@@ -0,0 +1,82 @@
...
+        timeout_min=slurm_options.TIMEOUT_MIN,
+        nodes=1,
+        cpus_per_task=slurm_options.CPUS_PER_TASK,
+        tasks_per_node=1,
+        mem_gb=slurm_options.MEM_GB,
+        slurm_additional_parameters=slurm_options.ADDITIONAL_PARAMETERS,
+    )
+
+    job = executor.submit(benchmark_suite_scheduler_job)
+    print(f"SUBMITTED EVALUATION JOB: {job.job_id}")
```
prigoyal, line 70: nit: use logger instead

tools/launch_benchmark_suite_scheduler_slurm.py
```diff
@@ -0,0 +1,82 @@
...
+
+    return job
```
prigoyal, line 72: question: do we need to return the job ?

tools/launch_benchmark_suite_scheduler_slurm.py
```diff
@@ -0,0 +1,82 @@
...
+    slurm_options = {**_DEFAULT_SLURM_OPTIONS, **override_slurm_options}
+    slurm_options = AttrDict(slurm_options)
+
+    executor.update_parameters(
+        name=slurm_options.NAME,
+        slurm_comment=slurm_options.COMMENT,
+        slurm_partition=slurm_options.PARTITION,
+        slurm_constraint=slurm_options.CONSTRAINT,
+        timeout_min=slurm_options.TIMEOUT_MIN,
+        nodes=1,
```
prigoyal, line 62: nit: I think we should make nodes a default option and use that setting. We should allow people to override it

tools/launch_benchmark_suite_scheduler_slurm.py
```diff
@@ -0,0 +1,82 @@
...
+        cpus_per_task=slurm_options.CPUS_PER_TASK,
+        tasks_per_node=1,
```
prigoyal, line 64: question: do we need to impose this constraint? aka can use configure this and override too?

tools/launch_benchmark_suite_scheduler_slurm.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+class SlurmEvaluatorJob:
+    def __init__(self, benchmark_suite_scheduler: BenchmarkSuiteScheduler):
+        self.benchmark_suite_scheduler = benchmark_suite_scheduler
+
+    def __call__(self):
+        self.benchmark_suite_scheduler.evaluate()
+
+    def checkpoint(self):
```
prigoyal, line 30: nit / question: it's a bit unclear what this function is doing . Can you share what the purpose is? :)

vissl/utils/misc.py
```diff
@@ -308,3 +311,83 @@ def __enter__(self):
...
+                time.sleep(ndelay)
+                ntries -= 1
+                ndelay *= backoff
+
+        return func(*args, **kwargs)
+
+    return wrapper
+
+
+def flatten_dict(d, parent_key="", sep="_"):
```
prigoyal, line 385: nit: let's add the type hints to the inputs.

prigoyal, line 385: nit: can we add a docstring to the function on what it does.

prigoyal (commented):
thank you so much @iseessel :) this is really awesome and very impactful feature addition to VISSL. 

1st round review! :)

vissl/engines/benchmark_suite_scheduler.py
```diff
@@ -0,0 +1,545 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
```
prigoyal, line 1: question: is this scheduler supposed to run on "every gpu" ?

typically in `vissl/engines/`, we have included engines that run on every gpu. the `distributed_launcher` takes care of launching the engine on each gpu worker. 

If this doesn't run on all gpus, then move this to either `vissl/utils/` folder of within the `tools/` merge this with the new .py added as this can also facilitate better code readability (avoid many hops) :)

vissl/engines/benchmark_suite_scheduler.py
```diff
@@ -0,0 +1,545 @@
...
+
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
```
prigoyal, line 5: nit: recommend adding a docstring that explains what this does

vissl/engines/benchmark_suite_scheduler.py
```diff
@@ -0,0 +1,545 @@
...
+from vissl.utils.slurm import is_submitit_available
+
+
+# How many times to retry a slurm job submission.
+_NUM_SLURM_RETRIES = 5
+# How many seconds to sleep between iterations of the main loop.
+_SLEEP_TIME_SECONDS = 15
+# Slurm states marked as terminal. SlurmEvulator#evaluate will finish
+# once all jobs are in a terminal state.
+_SLURM_JOB_TERMINAL_STATES = [
```
prigoyal, line 31: nit: remove the `SLURM` from the name

vissl/engines/benchmark_suite_scheduler.py
```diff
@@ -0,0 +1,545 @@
...
+                                retry all failed jobs.
+        max_retries: (int, optional, default=3). Maximum number of retries.
+        """
+        self.evaluation_jobs_finished = set()
+
+        # Required Arguments
+        self.training_checkpoint_dir = training_checkpoint_dir
+        self.training_checkpoint_file = os.path.join(
+            self.training_checkpoint_dir, "train_config.yaml"
+        )
```
prigoyal, line 112: nit: add an assert for it. if the `train_config.yaml` is not found, that means user training didn't work. so we should exit instantly.

vissl/engines/benchmark_suite_scheduler.py
```diff
@@ -0,0 +1,545 @@
...
+        slurm_evaluator_checkpoint: (str, optional, default=None) String of
+                                    slurm_evaluator checkpoint directory.
+        retry_evaluation_job_ids: (List[int], optional, default=[]) List of job_ids to retry.
+        auto_retry_evaluations: (bool, optional, default=False) Whether or not to automatically
+                                retry all failed jobs.
+        max_retries: (int, optional, default=3). Maximum number of retries.
+        """
+        self.evaluation_jobs_finished = set()
+
+        # Required Arguments
```
prigoyal, line 108: for all the required arguments, we must perform some kind of validation. In the `__init__` , you can call a `self.validate()` for instance that can take care of this.

vissl/engines/benchmark_suite_scheduler.py
```diff
@@ -0,0 +1,545 @@
...
+    "NODE_FAIL",
+    "OUT_OF_MEMORY",
+    "REVOKED",
+    "SPECIAL_EXIT",
+    "STOPPED",
+    "SUSPENDED",
+    "TIMEOUT",
+]
+# Wait for the training checkpoint folder to be available for 1 hour.
+_TRAINING_CHECKPOINT_FILE_WAIT_SECONDS = 60 * 60
```
prigoyal, line 46: this is a bit tricky. I think we shouldn't enforce this strictly and allow to override. for instance, for our SEER model, sometimes checkpoints appear every 2 hours etc

vissl/engines/benchmark_suite_scheduler.py
```diff
@@ -0,0 +1,545 @@
...
+                    f"Training checkpoint folder still doesn't exist after:"
+                    f"{_TRAINING_CHECKPOINT_FILE_WAIT_SECONDS / 60} minutes"
+                )
+
+            if PathManager.exists(self.training_checkpoint_file):
+                # Load training yaml config.
+                self._load_training_config()
+                break
+
+            time.sleep(_SLEEP_TIME_SECONDS)
```
prigoyal, line 156: question: could this be same as `_TRAINING_CHECKPOINT_FILE_WAIT_SECONDS` ? aka why do we need separate `_SLEEP_TIME_SECONDS` ?

tools/launch_benchmark_suite_scheduler_slurm.py
iseessel: This is slurm specific, so I think it should have slurm in the name. We will have something analagous for internal, so it would be good to be able to separate..

vissl/engines/benchmark_suite_scheduler.py
```diff
@@ -0,0 +1,545 @@
...
+
+        self.save_evaluation_results()
```
prigoyal, line 158: question: it sounds like we haven't run any evaluation yet - at this stage in the workflow, the checkpoint is just discovered. What does the `save_evaluation_results()` do ? we might want to find a different name for it to make it intuitive?

vissl/engines/benchmark_suite_scheduler.py
```diff
@@ -0,0 +1,545 @@
...
+        # Checkpoint folder is now available. Continuously monitor the training checkpoints,
+        # launch evaluation jobs as needed, monitor their progress, and record their results.
+        while True:
+            self._evaluate_checkpoints()
+
+            self._check_evaluation_jobs()
+
+            # Break if no more checkpoints to evaluate
+            if self._finished():
+                print("Evaluations are finished")
```
prigoyal, line 169: nit: use logger

tools/launch_benchmark_suite_scheduler_slurm.py
iseessel: Should we also remove "PARTITION: "learnfair"" From defaults.yaml?

tools/launch_benchmark_suite_scheduler_slurm.py
```diff
@@ -0,0 +1,82 @@
...
+
+import submitit
+from fvcore.common.file_io import PathManager
+from vissl.config.attr_dict import AttrDict
+from vissl.engines.benchmark_suite_scheduler import BenchmarkSuiteScheduler
+from vissl.utils.io import load_file
+
+
+# Default slurm options to pass to the executor.
+_DEFAULT_SLURM_OPTIONS = {
```
iseessel, line 11: TODO: Make this overrideable.

vissl/engines/benchmark_suite_scheduler.py
```diff
@@ -0,0 +1,545 @@
...
+            and benchmark["slurm_state"] != "COMPLETED"
+            and self.auto_retry_evaluations
+            and benchmark["num_retries"] < self.max_retries
+        )
+
+    @retry(n_tries=_NUM_SLURM_RETRIES)
+    def _launch_slurm_job(self, args, config):
+        return launch_distributed_on_slurm(engine_name=args.engine_name, cfg=config)
+
+    def _write_json_file(self, data, file_name):
```
prigoyal, line 310: can we reuse this function https://github.com/facebookresearch/vissl/blob/master/vissl/utils/io.py#L68 instead?

tools/launch_benchmark_suite_scheduler_slurm.py
iseessel: I think we should definitely make this overrideable, but I still think we should provide a sensible default.

vissl/engines/benchmark_suite_scheduler.py
```diff
@@ -0,0 +1,545 @@
...
+        Path(evaluation_dir).mkdir(parents=True, exist_ok=True)
+        self._write_json_file(self.evaluation_results, parent_metrics_file)
+
+        # Upload each checkpoint's evaluations to child directories.
+        for checkpoint_str, benchmarks in self.evaluation_results.items():
+            child_metrics_dir = os.path.join(evaluation_dir, checkpoint_str)
+            child_metrics_file = os.path.join(
+                child_metrics_dir, "evaluation_metrics.json"
+            )
+            Path(child_metrics_dir).mkdir(parents=True, exist_ok=True)
```
prigoyal, line 190: hmm we can try to use the `PathManager` ? the `Path` needs to be checked / tested if it will work with manifold (assuming we want to reuse this scheduler scripts more generically besides slurm)

vissl/engines/benchmark_suite_scheduler.py
```diff
@@ -0,0 +1,545 @@
...
+            self._write_json_file(benchmarks, child_metrics_file)
+
+        print("Saved evaluation Results")
```
prigoyal, line 193: nit: logger

vissl/engines/benchmark_suite_scheduler.py
```diff
@@ -0,0 +1,545 @@
...
+            # Break if no more checkpoints to evaluate
+            if self._finished():
+                print("Evaluations are finished")
+                break
+
+            time.sleep(_SLEEP_TIME_SECONDS)
+
+    def save_evaluation_results(self):
+        """
+        Upload json file to the evaluation directory.
```
prigoyal, line 176: nit: sounds like this function is doing more than just this?

tools/launch_benchmark_suite_scheduler_slurm.py
iseessel: It's a pretty low resource job w/ not many memory requirements. Maybe I can bump up to 32gb just to be safe.

vissl/engines/benchmark_suite_scheduler.py
```diff
@@ -0,0 +1,545 @@
...
+            self._check_evaluation_jobs()
+
+            # Break if no more checkpoints to evaluate
+            if self._finished():
+                print("Evaluations are finished")
+                break
+
+            time.sleep(_SLEEP_TIME_SECONDS)
+
+    def save_evaluation_results(self):
```
prigoyal, line 174: nit: would it make sense to rename it something like "setup_evaluation_out_folders" etc something like this i.e use the word `setup` ?

vissl/engines/benchmark_suite_scheduler.py
```diff
@@ -0,0 +1,545 @@
...
+            )
+            Path(child_metrics_dir).mkdir(parents=True, exist_ok=True)
+            self._write_json_file(benchmarks, child_metrics_file)
+
+        print("Saved evaluation Results")
+
+    def evaluation_dir(self):
+        return os.path.join(self.training_checkpoint_dir, "evaluations")
+
+    def _load_training_config(self):
```
prigoyal, line 198: nit: so it sounds like this function is doing more than just loading the training config. we should rename it to make it clear

vissl/engines/benchmark_suite_scheduler.py
```diff
@@ -0,0 +1,545 @@
...
+        # Load training yaml config.
+        self.training_config = load_file(self.training_checkpoint_file)
+        self.training_config = AttrDict(self.training_config)
```
prigoyal, line 201: do you think we can try to leverage the vissl https://github.com/facebookresearch/vissl/blob/master/vissl/utils/hydra_config.py#L24 ?

tools/launch_benchmark_suite_scheduler_slurm.py
iseessel: I like having this here, I think it gives some built-in documentation of what sort of options you can add to SLURM. Additional_parameters could be helpful for something like email notifcation setup.

vissl/engines/benchmark_suite_scheduler.py
```diff
@@ -0,0 +1,545 @@
...
+        print(
+            f"Loaded training checkpoint config from: { self.training_checkpoint_file }"
+        )
+        # Build main training task in order to extract iteration info.
+        self.training_task = build_task(self.training_config)
```
prigoyal, line 206: just noting here: we discussed getting rid of the training iterations. :)

vissl/engines/benchmark_suite_scheduler.py
```diff
@@ -0,0 +1,545 @@
...
+
+            log = f"""
+                Launched Slurm Evaluation job.
+                job_id: { job.job_id }, num_retries: { benchmark["num_retries"] }
+                evaluation_name: { benchmark["evaluation_name"] }
+                checkpoint_str: { checkpoint_str }
+                state_prev: None, state_current: { job.state }
+            """
+
+            print(log)
```
prigoyal, line 293: nit: logger

vissl/engines/benchmark_suite_scheduler.py
```diff
@@ -0,0 +1,545 @@
...
+        """
+        Validate the class instance is valid.
+        """
+        assert is_hydra_available(), "Make sure to install hydra"
+
+        assert (
+            is_submitit_available()
+        ), "Please 'pip install submitit' to schedule jobs on SLURM"
+
+    def _validate_training_cfg(self):
```
prigoyal, line 221: nit: let's rename it to `_validate_evaluation_setup` ?

tools/launch_benchmark_suite_scheduler_slurm.py
iseessel: I would think that this job would only ever require 1 node, as it is just a single process.

vissl/engines/benchmark_suite_scheduler.py
```diff
@@ -0,0 +1,545 @@
...
+        print(
+            f"Loaded training checkpoint config from: { self.training_checkpoint_file }"
+        )
+        # Build main training task in order to extract iteration info.
+        self.training_task = build_task(self.training_config)
+        self.training_task.dataloaders = self.training_task.build_dataloaders(False)
+        self.evaluation_results = self._generate_initial_evaluation_results()
+        self._validate_training_cfg()
+
+    def _validate_class_instance(self):
```
prigoyal, line 211: question: it sounds like we are simply checking availability of libraries like hydra, submitit and nothing to do with the class itself. Could we find a better place for these functions and possible just call them directly instead of putting them under the `_validate_class_instance` ? :)

tools/launch_benchmark_suite_scheduler_slurm.py
iseessel: I think we should set this to 1. There will only be 1 task run here on SLURM. (If you set multiple tasks, it will run the same job multiple times with the same config).

tools/launch_benchmark_suite_scheduler_slurm.py
iseessel: nope, I'll remove!

prigoyal (commented):
2nd round :)

tools/launch_benchmark_suite_scheduler_slurm.py
prigoyal: that's fine with me :)

tools/launch_benchmark_suite_scheduler_slurm.py
prigoyal: It's highly likely that we specify constraints like: "volta32g" to be able to leverage 32GB machines for most of our evaluations.

tools/launch_benchmark_suite_scheduler_slurm.py
prigoyal: gotcha. this is something captured by the SLURM config already. Do we require a new description still? :)

tools/launch_benchmark_suite_scheduler_slurm.py
prigoyal: gotcha. For quite a few evals, it's the case that we use > 1 node as some jobs take long time . I think we need to configure this.

iseessel: @prigoyal Ah, I see the confusion. This slurm configuration is for the evaluation job, *not* the evaluations. The evaluations config are set in #launch_distributed_slurm.

vissl/engines/benchmark_suite_scheduler.py
iseessel: So I have set it up, so that this can be launched simultaneously when the training is launched. Since we don't know when each job will be executed by SLURM (and likely the low resource job will be executed first), I wait for the training config to become available for a set amount of time, then fail.

vissl/engines/benchmark_suite_scheduler.py
iseessel: This is SLURM specific, this will all be refactored shortly to take into account for fbinfra.

vissl/engines/benchmark_suite_scheduler.py
iseessel: Yeah agreed, this is a mistake -- we should use hydra_config.

vissl/engines/benchmark_suite_scheduler.py
iseessel: So we still need the training iterations, we just wanted to find them in a different way.

I synced up with @QuentinDuval on this. I went down the path of Introducing them dynamically, but this complicates the logic too much. I much prefer scaffolding them at the beginning, it's much easier to follow and debug.

configs/config/benchmark/suite/benchmark_suite_scheduler_example.json
```diff
@@ -0,0 +1,22 @@
+{
+ "params": {
+        "training_checkpoint_dir": "(str) Training checkpoint directory. That is the CHECKPOINT.DIR of the training config",
```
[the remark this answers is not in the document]
iseessel: Agree we should add something to README -- since approach will change shortly with added fbinfra support, I'd like to wait on that.

tools/launch_benchmark_suite_scheduler_slurm.py
iseessel: See https://github.com/facebookresearch/vissl/pull/325/files#r645636434

tools/launch_benchmark_suite_scheduler_slurm.py
iseessel: This is for the eval monitoring script, not the evaluations.

tools/launch_benchmark_suite_scheduler_slurm.py
iseessel: I don't think I understand what you mean. We are not using the defaults.yaml here, we are providing our own default options here (mainly so that we can set the CPUS_PER_TASK and MEM_GB intelligently.

vissl/engines/benchmark_suite_scheduler.py
iseessel: Ah this is confusingly named. This is the amount of time we are waiting for the training folder to be present, in order to read it's YAML config.

vissl/engines/benchmark_suite_scheduler.py
iseessel: These are two separate concepts. _TRAINING_CHECKPOINT_FILE_WAIT_SECONDS is to wait for the training config YAML to be available (I'm changing name to make it clearer). 

_SLEEP_TIME_SECONDS is just to sleep a small amount of time in order to not constantly be checking for checkpoints, since checkpoints don't occur frequently and I want to not overwhelm submitit with requests for job status.

vissl/engines/benchmark_suite_scheduler.py
iseessel: So the reason I used print and not logger was because I wasn't seeing the stdout when running it via CLI when using logger. I'll change though.

vissl/engines/benchmark_suite_scheduler.py
iseessel: No we can't. I tried this initially, but this function assumes a list of dicts and writes out the json in newline_json_delimited form.

tools/launch_benchmark_suite_scheduler_slurm.py
iseessel: We should use print here, logging.info doesn't display to stdout screen when using CLI.

tools/launch_benchmark_suite_scheduler_slurm.py
iseessel: TODO: Create defaults json file.

tools/benchmark_suite_scheduler.py
```diff
@@ -0,0 +1,582 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
```
prigoyal, line 1: nit: as discussed, let's move this to under `vissl/utils/`

tools/benchmark_suite_scheduler.py
```diff
@@ -0,0 +1,582 @@
...
+
+"""
+    This class is designed to be used to run multiple evaluations on a single (pre)training.
+    Using the #evaluate method we continuously monitor training checkpoints, launch evaluations
+    dynamically as they become available, and amalgamate the evaluation results as they become
+    available.
+
+    For SLURM usage, you should create a JSON configuration file (see benchmark_suite_scheduler_template.json)
+    and use the launch_benchmark_suite_scheduler_slurm.sh for convenience.
+"""
```
prigoyal, line 33: nit: no indent

tools/launch_benchmark_suite_scheduler_slurm.py
```diff
@@ -0,0 +1,109 @@
...
+from fvcore.common.file_io import PathManager
+from tools.benchmark_suite_scheduler import BenchmarkSuiteScheduler
+from vissl.config.attr_dict import AttrDict
+from vissl.utils.hydra_config import is_hydra_available
+from vissl.utils.io import load_file
+from vissl.utils.slurm import is_submitit_available
+
+
+# Default slurm options to pass to the executor.
+_DEFAULT_SLURM_OPTIONS = {
```
prigoyal, line 19: as discussed, let's move this to dedicated settings in defaults.yaml :)

tools/launch_benchmark_suite_scheduler_slurm.py
```diff
@@ -0,0 +1,109 @@
+#!/bin/bash
```
prigoyal, line 1: should this be removed?

tools/benchmark_suite_scheduler.py
```diff
@@ -0,0 +1,582 @@
...
+            and benchmark["slurm_state"] != "COMPLETED"
+            and self.auto_retry_evaluations
+            and benchmark["num_retries"] < self.max_retries
+        )
+
+    @retry(n_tries=_NUM_SLURM_RETRIES)
+    def _launch_slurm_job(self, args, config):
+        return launch_distributed_on_slurm(engine_name=args.engine_name, cfg=config)
+
+    def _write_json_file(self, data, file_name):
```
prigoyal, line 347: nit: can we use the function `save_file` https://github.com/facebookresearch/vissl/blob/master/vissl/utils/io.py#L68 which does exactly this?

tools/benchmark_suite_scheduler.py
```diff
@@ -0,0 +1,582 @@
...
+        Create the /evaluations directory inside the training checkpoints dir.
+        Upload json file to the parent evaluation directories, as well as
+        to each child evaluation directories.
+        """
+        # Upload all checkpoints evaluations to parent checkpoint directory.
+        evaluation_dir = self.evaluation_dir()
+        parent_metrics_file = os.path.join(evaluation_dir, "evaluation_metrics.json")
+
+        if not PathManager.exists(evaluation_dir):
+            PathManager.mkdirs(evaluation_dir)
```
prigoyal, line 207: nit: can we leverage this function directly https://github.com/facebookresearch/vissl/blob/master/vissl/utils/io.py#L122-L133 ?

tools/benchmark_suite_scheduler.py
```diff
@@ -0,0 +1,582 @@
...
+
+        # Upload each checkpoint's evaluations to child directories.
+        for checkpoint_str, benchmarks in self.evaluation_results.items():
+            child_metrics_dir = os.path.join(evaluation_dir, checkpoint_str)
+            child_metrics_file = os.path.join(
+                child_metrics_dir, "evaluation_metrics.json"
+            )
+
+            if not PathManager.exists(child_metrics_dir):
+                PathManager.mkdirs(child_metrics_dir)
```
prigoyal, line 219: nit: can we use this function directly: https://github.com/facebookresearch/vissl/blob/master/vissl/utils/io.py#L122-L133 ?

prigoyal (commented):
looks great! thank you so much . Some inline comments and also take a look at some open conversations/resolve them.

**Remaining steps:**
- [ ] Update the test plan with the ongoing training + evals
- [ ] Update the test plan with the static evals with a checkpoint from model zoo. Pick the checkpoint for SwAV from here https://github.com/facebookresearch/vissl/blob/master/MODEL_ZOO.md#SwAV and the json is here https://github.com/facebookresearch/vissl/blob/master/configs/config/model_zoo/benchmark_in1k_linear_deepclusterv2_swav.json#L66-L83 . Create a dev/benchmark_suite/example_swav.json for instance if needed
- [ ] rebase and import to phabricator again . Ensure all tests pass (internal, external)
- [ ] We will want to post a workplace post afterwards. We will want to share the results visually from the test plan so any figures, snapshots you can grab , do it :)

tools/benchmark_suite_scheduler.py
iseessel: I responded to this earlier, but because of the file changes, the comment got buried. 

They do slightly different things. 
1. #save_file appends, whereas this overwrites. 
2. #save_file adds a new_line.

tools/benchmark_suite_scheduler.py
iseessel: Nice. Thanks, I missed this one!

dev/benchmark_suite/benchmark_suite_scheduler_swav.json
```diff
@@ -0,0 +1,25 @@
+{
```
iseessel, line 1: Revert this, this is for testing.

tools/launch_benchmark_suite_scheduler_slurm.py
```diff
@@ -0,0 +1,106 @@
...
+    )
+    executor = submitit.AutoExecutor(folder=benchmark_suite_scheduler.evaluation_dir())
+
+    assert "slurm_options" in config, "slurm_options must be specified"
+    assert (
+        "PARTITION" in config["slurm_options"]
+    ), "slurm_options.PARTITION is a required field to launch the benchmark suite on slurm"
+
+    slurm_options = AttrDict(config["slurm_options"])
+    benchmark_suite_scheduler.evaluate()
```
iseessel, line 75: This is for testing purposes, revert once finished testing.

prigoyal:
looks great. Ready to merge. Thank you so much @iseessel  and thank you @QuentinDuval for amazing mentoring on this. 

@iseessel , this PR needs a rebase + reimport and then ready to merge.

PR #333, opened 2021-06-04
Move hydra dict conversions before saving dict to disk

iseessel:
We should be saving the final config file on disk that is used by training and not have any changes after it's saved.


iseessel:
Seems like this broke a test, a bit confused why. Will look at soon.

PR #334, opened 2021-06-04
Remove default slurm partition option

iseessel:
I am a bit worried this may break somebody's workflow? We might want to have our own internal defaults in order to ensure backwards compatability.


prigoyal (approved):
makes sense to me!

QuentinDuval:
Make sense to me too. +1 for merging it !

prigoyal:
this needs to be rebased and then reimported to phabricator to land :)

iseessel:
> this needs to be rebased and then reimported to phabricator to land :)

Done! @prigoyal

PR #338, opened 2021-06-10
Fix seed regression, as option is not available in prev pytorch version

iseessel:
Our recommended pytorch version does not have seed available as an option for `torch.utils.data.distributed.DistributedSampler`. This was breaking master for me in distributed training on SLURM.


prigoyal:
looks good to me.

Issue #342, opened 2021-06-14
JSON file format changes due to appending


pranavsinghps1:
## JSON file format changes due to appending
 Re-running the 
```
from vissl.utils.io import save_file
save_file(json_data, "/content/configs/config/dataset_catalog.json")
``` 

appends to the existing JSON file instead of re-writing it, this leads to a JSONDecoderError, which means the JSON file is not of the appropriate format.

## Expected behavior:

Changing the
 ```save_file(json_data, "/content/configs/config/dataset_catalog.json")``` to something that re-writes it instead of appending would help solve the problem.

Right now to fix it the file (dataset_catalog.json) needs to be deleted before re-running the above code snippet.

## Environment:

I used google Colab to run the jupyter notebooks: 
https://github.com/facebookresearch/vissl/blob/master/tutorials/Understanding_VISSL_Training_and_YAML_Config.ipynb
 and also in https://github.com/facebookresearch/vissl/blob/master/tutorials/Benchmark_Full_Finetuning_on_ImageNet_1K.ipynb

prigoyal:
Hi @pranavsinghps1 , thank you for reporting this. Indeed, I think we should support optional user input on whether to re-write and append i.e. something like `save_file(json_data, "my_file.json", rewrite=False)`  so user can configure and control this.

iseessel:
Fixed by: https://github.com/facebookresearch/vissl/pull/346

PR #344, opened 2021-06-17
JSONDecoderError Fix

pranavsinghps1:
With respect to #342, re-running the cell won't cause JSONDecoderError.

### DIFF
--- a/vissl/utils/io.py
+++ b/vissl/utils/io.py
@@ -66,7 +66,7 @@ def save_file(data, filename):
         with PathManager.open(filename, "wb") as fopen:
             np.save(fopen, data)
     elif file_ext == ".json":
-        with PathManager.open(filename, "a") as fopen:
+        with PathManager.open(filename, "w") as fopen:
             fopen.write(json.dumps(data, sort_keys=True) + "\n")
             fopen.flush()
     elif file_ext == ".yaml":


vissl/utils/io.py
```diff
@@ -66,7 +66,7 @@ def save_file(data, filename):
         with PathManager.open(filename, "wb") as fopen:
             np.save(fopen, data)
     elif file_ext == ".json":
-        with PathManager.open(filename, "a") as fopen:
+        with PathManager.open(filename, "w") as fopen:
```
iseessel, line 69: Thanks so much Pranav for submitting a PR.

The method save_file is used by other parts of VISSL, and as such we need to make sure it works with the other parts. I would suggest the following: 

1. Add option `append`, with default False to #save_file.
2. Add in code to support json append=True and append=False. 
3. Change [here](https://github.com/facebookresearch/vissl/blob/4b19ab28953470d615a25273fa22eab659528867/vissl/hooks/log_hooks.py#L498)) to use append=True. This is the only use of #save_file in the codebase using json format. Global search save_file to validate this.

CC: @prigoyal

pranavsinghps1: Continued at #346

iseessel:
Closing in favor of:  #346

PR #345, opened 2021-06-17
[vissl] Fix multicrop transform total_num_crop overwriting

iseessel:


vissl/utils/hydra_config.py
```diff
@@ -338,7 +337,9 @@ def infer_losses_config(cfg):
         cfg.LOSS.multicrop_simclr_info_nce_loss.buffer_params.effective_batch_size = (
             batch_size * world_size
         )
-        cfg.LOSS.multicrop_simclr_info_nce_loss.num_crops = total_num_crops
+        cfg.LOSS.multicrop_simclr_info_nce_loss.num_crops = (
+            total_num_crops or cfg.LOSS.multicrop_simclr_info_nce_loss.num_crops
+        )
```
prigoyal, line 342: I feel that we should just remove the computation of `total_num_crops`. From a pure reading perspective, IMO, its adding complexity that outweights the benefits. In some places, it seems not essential either.

prigoyal (commented):
thank you @iseessel , really appreciate your help with this. I left an inline comment and I propose to shoot for simplicity.

vissl/utils/hydra_config.py
iseessel: Hmm... I do agree it's not the most readable, but I do prefer keeping it. 

I see a lot of configs where total_num_crops != num_crops (we could of course change them), but I feel like copy and pasting configs is common and inferring this prevents against someone making a mistake when dealing with the config.

Another idea is we could reverse how we are inferring -- that is we could infer the "transformation.total_num_crops" from the `loss.num_crops`.

prigoyal: gotcha. let's chat offline about this and find a plan. :)

PR #363, opened 2021-07-16
Updated example in img_pil_to_multicrop.py docstring

growlix:
Previous docstring example: total_num_crops=2, num_crops=[1, 2]
Updated: total_num_crops=3, num_crops=[1, 2]


iseessel (approved):
Thanks for catching.

Issue #364, opened 2021-07-17
Extract_cluster from DeepCluster_V2 pretrained model


DC95:
Hello @QuentinDuval and @prigoyal 

I was wondering whether I can use the extract_cluster from configs/config/extract_cluster/swav/visualise_swav_resnet_in1k_8gpu.yaml for **deep_clusterV2**  pretrained model?

I tried with one of the basic changes with the **head configuration** i.e. I tried the one which I used in my pretraining- 

    HEAD:
       PARAMS: [
         ["mlp", {"dims": [2048, 2048], "use_relu": True, "skip_last_layer_relu_bn": False}],
         ["mlp", {"dims": [2048, 128]}],
       ]
      
But it came up with the error - 

> File "/p/project/deepacf/kiste/DC/vissl/vissl/trainer/trainer_main.py", line 514, in extract_clusters
    assert self.task.base_model.is_clustering_model(), error_message
AssertionError: Extracting clusters is only available for pre-training methods based on clusters

**My understanding -** 

I looked at the code from **trainer_main.py** and the function **def extract_clusters(self)**,  here it is mentioned - 

_extract multi-GPU cluster extraction for pre-trained models based on clusterization (SwAV, **DeepCluster**, etc.)_

But I also went to function **def [is_clustering_model**](https://github.com/facebookresearch/vissl/blob/6c21cd91e66dcf74f4533463977514bb11a2cd1e/vissl/models/base_ssl_model.py#L551) (used for asserting the base model line 514 in def extract_clusters) where it is written -

_- only supported for **SWAV currently**_ (This contradicts the above)

I think extract_cluster is not yet ready for deepcluster_v2. Can somehow we do it??

QuentinDuval:
Hi @DC95,

Indeed, the extract cluster is for the moment only working for SwAV and the message is a bit misleading.

I think we can schedule an enhancement on our side to improve this and support DeepClusterV2 in addition to SwAV (@prigoyal what do you think?)?

Or if you feel comfortable trying to contribute it (it seems you went through some of the code already), give it a go and open a PR with it: we will very much welcome your help :)

DC95:
Dear @QuentinDuval and @prigoyal 

I have not done this before :) but I will give it a go. I will definitely need your support. 

Regards,
DC

DC95:
Dear @QuentinDuval 

From today, I have started to look deeper into the codes and management of VISSL. But I see that you have self-assigned it yourself. 

[...]

PR #417, opened 2021-09-05
Extract_cluster from DeepCluster_V2 pretrained model and visualisation (#364)

DC95:
Summary:
# For DC_V2 Cluster_assignment_extraction + visualisation_yaml_config

This PR introduces:
1. Changes in the trainer_main.py and base_ssl_model.py to adjust the head configuration of DC_V2
2. Yaml file config to extract clusters from the DC_V2 train checkpoint.

My own doubts - 

Forgive me for my mistakes; I don't feel confident whether changing - 

1. prototype_index[idx].item() **to** prototype_index.item() is doing the correct job in base_ssl_model.py. I changed it because it was otherwise showing the error - _'IndexError: invalid index of a 0-dim tensor. Use tensor.item() to convert a 0-dim tensor to a Python number'_
2. I also understand that in DC_V2, we use multi clustering, and for that we there are multi heads (i.e., len(self.heads) >1).

This is as far as my understanding goes. Kindly improve it as I am in urgent need of the visualization of the DCV2 Cluster assignments.

Regards,
DC

Kind note - 
Rest we can proceed with cluster_assignments_to_dataset.py as given in the [link](https://github.com/facebookresearch/vissl/commit/a8622d7649e9392e13db4806e014f4048678e9ba)

## Interested parties

CC: @QuentinDuval, @iseessel

### DIFF
--- a/vissl/models/base_ssl_model.py
+++ b/vissl/models/base_ssl_model.py
@@ -13,7 +13,7 @@ from classy_vision.models import ClassyModel, register_model
 from fairscale.nn.data_parallel import FullyShardedDataParallel as FSDP
 from vissl.config import AttrDict
 from vissl.data.collators.collator_helper import MultiDimensionalTensor
-from vissl.models.heads import SwAVPrototypesHead, get_model_head
+from vissl.models.heads import SwAVPrototypesHead, get_model_head, MLP
 from vissl.models.model_helpers import (
     get_trunk_output_feature_names,
     is_feature_extractor_model,
@@ -551,11 +551,11 @@ class BaseSSLMultiInputOutputModel(ClassyModel):
     def is_clustering_model(self):
         """
         Whether or not the model is a clustering model
-        (only supports SwAV for the moment)
+        (supports SwAV and DeepCluster_v2 for the moment)
         """
-        return len(self.heads) == 1 and any(
-            isinstance(module, SwAVPrototypesHead) for module in self.heads[0].modules()
-        )
+        return len(self.heads) == 1 or any(
+            isinstance(module, MLP) for module in self.heads[0].modules()) or any(
+            isinstance(module, SwAVPrototypesHead) for module in self.heads[0].modules())
 
     @property
     def num_classes(self):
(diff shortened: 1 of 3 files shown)


vissl/models/base_ssl_model.py
```diff
@@ -551,11 +551,11 @@ def init_model_from_weights_params_file(
...
         """
         Whether or not the model is a clustering model
-        (only supports SwAV for the moment)
+        (supports SwAV and DeepCluster_v2 for the moment)
         """
-        return len(self.heads) == 1 and any(
-            isinstance(module, SwAVPrototypesHead) for module in self.heads[0].modules()
-        )
+        return len(self.heads) == 1 or any(
+            isinstance(module, MLP) for module in self.heads[0].modules()) or any(
```
iseessel, line 557: So I don't think any model with an MLP head should return true for `is_clustering_model`. 

Since it's difficult to find a way to tell whether or not the model is DeepCluster in `base_ssl_model.py` -- I would recommend moving #is_clustering_model to SelfSupervisionTask `train_task.py`. 

So `trainer_main.py` reads: 

```
        error_message = "Extracting clusters is only available for pre-training methods based on clusters"
        assert self.task.is_clustering_model(), error_message
```

You can then use `task.loss.loss_config['name'] == 'deepclusterv2_loss'` as a requirement. LMK if that makes sense.

iseessel (changes requested):
Thanks for the PR and glad you figured out how to extract the clusters for deep_clusterv2 :). I have some comments/concerns about PR.

vissl/models/base_ssl_model.py
```diff
@@ -551,11 +551,11 @@ def init_model_from_weights_params_file(
     def is_clustering_model(self):
         """
         Whether or not the model is a clustering model
-        (only supports SwAV for the moment)
+        (supports SwAV and DeepCluster_v2 for the moment)
         """
-        return len(self.heads) == 1 and any(
-            isinstance(module, SwAVPrototypesHead) for module in self.heads[0].modules()
-        )
+        return len(self.heads) == 1 or any(
```
iseessel, line 556: This change doesn't really make sense to me -- can you help explain?

1. This doesn't look like it would be backwards compatible when using the SwAVPrototypesHead.
2. Here you are setting the cluster_assignment for different images to the same value?

For deepclusterv2, the clustering ocurrs in the loss. See `deepclusterv2_loss.py`.

iseessel (changes requested):

iseessel:
Is this PR still needed? @DC95 I know we resolved the issue of saving the deepcluster clusters.

PR #442, opened 2021-10-07
Add augly transformation support

iseessel:
Differential Revision: D31462923

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,7 +16,7 @@ cpu: &cpu
   environment:
     TERM: xterm
   machine:
-    image: default
+    image: ubuntu-1604:201903-01
   resource_class: medium
 
 gpu: &gpu
@@ -37,8 +37,8 @@ install_python: &install_python
       working_directory: ~/
       command: |
         pyenv versions
-        pyenv install 3.6.2
-        pyenv global 3.6.2
+        pyenv install -f 3.7.0
+        pyenv global 3.7.0
 
 update_gcc7: &update_gcc7
   - run:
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -153,17 +165,18 @@ jobs:
       # Cache the vissl_venv directory that contains dependencies
       - restore_cache:
           keys:
-            - v5-cpu-dependencies-{{ checksum "requirements.txt" }}-{{ checksum "setup.py" }}
+            - v8-cpu-dependencies-{{ checksum "requirements.txt" }}-{{ checksum "setup.py" }}
```
prigoyal, line 168: question: why bumping from v5 to v8 directly? why not v6?

.circleci/config.yml
```diff
@@ -16,7 +16,7 @@ cpu: &cpu
   environment:
     TERM: xterm
   machine:
-    image: default
+    image: ubuntu-1604:201903-01
```
prigoyal, line 19: question: why are we changing the image to ubuntu? why not use the default container

tests/test_transforms.py
```diff
@@ -77,3 +82,16 @@ def test_img_pil_to_multicrop(self):
...
             self.assertEqual((96, 96), crop.size)
+
+    def test_augly_transforms(self):
+        cfg = compose_hydra_configuration(
+            [
+                "config=test/cpu_test/test_cpu_resnet_simclr.yaml",
+                "+config/test/cpu_test/transforms=augly_transforms_example",
+            ],
+        )
+        args, config = convert_to_attrdict(cfg)
```
prigoyal, line 93: double check: make sure these tests don't request gpus.

vissl/utils/misc.py
```diff
@@ -80,6 +80,19 @@ def is_apex_available():
     return apex_available
 
 
+def is_augly_available():
```
prigoyal, line 83: could we also assert somewhere about the python version if augly has a specific requirement?

prigoyal (commented):
thank you @iseessel , added some thoughts inline.

